PR #8, opened 2024-06-03
Only set vip for controlplane nodes

### LINKED ISSUE #7: Node `worker` template has `floatingIP` `vip` section
https://github.com/aenix-io/talm/blob/71dacecb82be39f98a95fc5a569a47926fc7bf5a/charts/cozystack/templates/_helpers.tpl#L42C7-L45C17

This part of the talos.config template makes the `worker` config invalid when a VIP is used.

I'm not sure exactly how to fix it in the template itself, but the error I got from talm CLI was suggestive pretty helpfully that removing this section from the worker node definition looked to be indicated, and that worked.

Just starting to use `talm` after wrapping my head around the talos-bootstrap scripts! Thanks for making this, it's great :) 😊

kingdonb:
Built talm locally and tested this out, it should fix #7

### DIFF
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -39,10 +39,10 @@ machine:
       routes:
         - network: 0.0.0.0/0
           gateway: {{ include "talm.discovered.default_gateway" . }}
-      {{- with .Values.floatingIP }}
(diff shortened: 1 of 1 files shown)


kvaps:
Thanks for your contribution, fixed in `v0.5.7`

PR #37, opened 2025-03-12
Add an option to support Nvidia GPUs.

xy2:
The Nvidia kernel modules try to check for hardware when they load into the kernel. If there is no GPU actually present, dmesg will be full of error messages. Make these modules load only if the setting is explicitly true.

Also, dropped the containerd 1.0 config part. The last Talos version that contains it is 1.7.7

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced NVIDIA GPU support that automatically applies tailored configuration settings when an NVIDIA GPU is detected. This includes optimized driver and runtime adjustments as well as improved system security controls for GPU-enabled environments.
  - Added a template variable to recognize the presence of an NVIDIA GPU, enhancing configuration flexibility.

### DIFF
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -20,11 +20,21 @@ machine:
         - usermode_helper=disabled
     - name: zfs
     - name: spl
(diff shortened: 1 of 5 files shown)


kvaps:
Closed in favor https://github.com/cozystack/talm/pull/46

Issue #34, opened 2025-02-14
Error: either install disk or diskSelector should be defined


keynight:
Logs 

`
tail --version
tail (GNU coreutils) 8.32
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Paul Rubin, David MacKenzie, Ian Lance Taylor,
and Jim Meyering.

talm init --preset cozystack
Created secrets.yaml
generating PKI and tokens
Created talosconfig
Created charts/talm/Chart.yaml
Created Chart.yaml
Created templates/_helpers.tpl
Created values.yaml
Created charts/talm/templates/_helpers.tpl
Created templates/controlplane.yaml
Created templates/worker.yaml

mkdir nodes

talm template -e 192.168.100.110 -n 192.168.100.110 -t templates/controlplane.yaml -i > nodes/srv110.yaml

talm apply -f nodes/srv110.yaml -i
- talm: file=nodes/srv110.yaml, nodes=[192.168.100.110], endpoints=[192.168.100.110]
error applying new configuration: rpc error: code = InvalidArgument desc = configuration validation failed: 1 error occurred:
        * either install disk or diskSelector should be defined
        
`

![Image](https://github.com/user-attachments/assets/e525ed1c-f452-4aac-b138-059b5fe9dfe6)

keynight:
Workaround: add  "disk: /dev/sda" to your config yaml

![Image](https://github.com/user-attachments/assets/5d5e10e6-e12a-46fd-88e4-32930e34bb8b)

chumkaska:
> srv110.yaml

Show the disk section in srv110.yaml, it might not have been detected when creating the template.

spa086:
Have some issue.

```
# talm: nodes=["192.168.100.16"], endpoints=["192.168.100.16"], templates=["templates/controlplane.yaml"]
# THIS FILE IS AUTOGENERATED. DO NOT EDIT IT!
machine:
  type: controlplane
  kubelet:
    nodeIP:
      validSubnets:
        - 192.168.100.0/24
  network:
    hostname: talos-a0031
    # -- Discovered interfaces:
    # enxbc2411aed34b:
    #   id: ens18
    #   hardwareAddr:bc:24:11:ae:d3:4b
    #   busPath: 0000:00:12.0
[...]

PR #41, opened 2025-03-30
Fix Error: either install disk or diskSelector should be defined #34

keynight:
fix: Default to /dev/sda if no disk is found

Updated the disk selection logic to use /dev/sda as the default value when no disk is discovered. This ensures that the installation process has a fallback option, improving reliability in environments where disks may not be detected.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the disk configuration process by automatically falling back to a default disk when none is detected, ensuring consistent and robust system behavior.

### DIFF
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -34,7 +34,12 @@ machine:
     image: {{ . }}
     {{- end }}
     {{- (include "talm.discovered.disks_info" .) | nindent 4 }}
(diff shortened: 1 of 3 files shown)


keynight:
Error: either install disk or diskSelector should be defined #34

keynight:
Error: either install disk or diskSelector should be defined #34

kvaps (approved):
Thank you @keynight I fixed style a little bit and replicated this to the other presets

kvaps:
Thank you, accepted to v0.9.0

PR #119, opened 2026-04-10
fix(commands): render templates online in apply to resolve lookups

myasnikovdaniil:
## Summary

- `talm apply` with modeline `templates=` was rendering templates **offline** (no Talos client), so all `lookup()` calls returned empty values — producing configs with empty `interface` names
- Talos v1.12 now rejects these with: `[networking.os.device.interface], [networking.os.device.deviceSelector]: required either config section to be set`
- Fix: create the Talos client **before** rendering and pass it to `engine.Render` with `Offline=false`, so `lookup()` functions resolve real discovery data from the target node
- This makes `talm apply` a true single-step "render + apply" command, matching the pattern already used by `talm template`

### What changed in `apply.go`

**Before** (broken flow):
```
render offline (no client, lookups empty) → create client → apply broken config
```

**After** (fixed flow):
```
create client → render online (with client, lookups work) → apply valid config
```

Extracted inline `withClient` closure into a named `withApplyClient` helper. Both the template rendering path and the direct patch path now create the client via this helper. For the template path, `engine.Render` is called **inside** the client callback with the real client.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./pkg/commands/ -run TestBuildApply -v` passes (updated assertion: `Offline=false`)
- [x] `go test ./pkg/... -v` — all tests pass
- [ ] Manual: `talm apply -f nodes/master1.yaml -i` renders templates online with real lookups and applies to Talos v1.12 node

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Template rendering during configuration apply now runs with a live connection so discovery-dependent templates resolve correctly.
  * Client selection and initialization centralized for consistent behavior across modes; apply results are printed after application.
  * Error messages clarified to provide better context when rendering or applying configurations fails.
* **Tests**
  * Updated expectations for online rendering; added tests covering lookup behavior in offline vs online rendering.

### DIFF
--- a/pkg/commands/apply.go
+++ b/pkg/commands/apply.go
@@ -132,7 +132,7 @@ func apply(args []string) error {
 					TryModeTimeout: durationpb.New(applyCmdFlags.configTryTimeout),
 				})
 				if err != nil {
-					return fmt.Errorf("error applying new configuration: %s", err)
+					return fmt.Errorf("error applying new configuration: %w", err)
 				}
 
 				helpers.PrintApplyResults(resp)
@@ -153,7 +153,7 @@ func apply(args []string) error {
 				return fmt.Errorf("error serializing configuration: %w", err)
 			}
 
-			err = withApplyClient(func(ctx context.Context, c *client.Client) error {
+			if err := withApplyClient(func(ctx context.Context, c *client.Client) error {
 				fmt.Printf("- talm: file=%s, nodes=%s, endpoints=%s\n", configFile, GlobalArgs.Nodes, GlobalArgs.Endpoints)
 
 				resp, err := c.ApplyConfiguration(ctx, &machineapi.ApplyConfigurationRequest{
@@ -163,13 +163,15 @@ func apply(args []string) error {
 					TryModeTimeout: durationpb.New(applyCmdFlags.configTryTimeout),
 				})
 				if err != nil {
-					return fmt.Errorf("error applying new configuration: %s", err)
+					return fmt.Errorf("error applying new configuration: %w", err)
 				}
 
 				helpers.PrintApplyResults(resp)
 
 				return nil
-			})
+			}); err != nil {
+				return err
+			}
 		}
 		if err != nil {
 			return err
--- a/pkg/commands/apply.go
+++ b/pkg/commands/apply.go
@@ -112,14 +112,33 @@ func apply(args []string) error {
 		// Resolve secrets.yaml path relative to project root if not absolute
 		withSecretsPath := ResolveSecretsPath(applyCmdFlags.withSecrets)
 
-		var result []byte
 		if len(modelineTemplates) > 0 {
-			// Template rendering path: render templates via Helm engine to produce full config
+			// Template rendering path: connect to the node first, render templates
+			// online (so lookup() functions resolve real discovery data), then apply.
 			opts := buildApplyRenderOptions(modelineTemplates, withSecretsPath)
-			result, err = engine.Render(ctx, nil, opts)
-			if err != nil {
-				return fmt.Errorf("template rendering error: %w", err)
-			}
+
+			err = withApplyClient(func(ctx context.Context, c *client.Client) error {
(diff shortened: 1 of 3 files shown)


lexfrei (changes requested):
## Requesting changes

The fix itself is correct — moving rendering inside the client callback so `lookup()` functions resolve real discovery data is the right approach, and the shadow/wrap bugs flagged by the bots have been properly addressed. However, the PR is missing the most important artifact: **a test that proves the bug from the reports**.

### Required: regression test for the actual bug

The test plan only flips `Offline=true` → `Offline=false` in `TestBuildApplyRenderOptions`. That verifies a struct field value, not behavior. The real bug was:

1. `apply` renders templates offline → `lookup()` returns empty maps
2. Helpers like `talm.discovered.default_link_name_by_gateway` produce empty strings
3. Rendered config has `interface: ""` (or missing)
4. Talos 1.12 rejects it with `[networking.os.device.interface], [networking.os.device.deviceSelector]: required either config section to be set`

Please add a regression test that exercises this path end-to-end. Suggested shape:

- Use a fake `LookupFunc` that returns distinguishable values online vs empty offline (the codebase already swaps `helmEngine.LookupFunc` in existing render tests — see `pkg/engine/render_test.go`)
- Render a minimal chart template that uses `talm.discovered.default_link_name_by_gateway` (or any helper that consumes `lookup`)
- Assert: with `Offline=true`, the rendered config has an empty interface field (reproduces the bug)
- Assert: with `Offline=false` + the fake lookup, the rendered config has a populated interface field (proves the fix)
- Bonus: run the output through `isTalosConfigPatch` + Talos config validation to show the exact error the user reported surfaces only in the broken path

Without this test, nothing prevents someone from flipping `Offline` back to `true` in a future refactor and silently reintroducing the regression.

### Bonus action items (not blocking, happy to track as separate issues)

1. **Multi-node regression**: `engine.Render` now calls `helpers.FailIfMultiNodes` in non-insecure mode. If a user puts multiple IPs in a single modeline `nodes=[ip1, ip2]` or passes `--nodes ip1,ip2`, apply will now fail before rendering. The typical single-node-per-file flow from the README still works, and maintenance mode (`-i`) is unaffected, but the behavior change is worth documenting or rejecting explicitly with a clearer message.

2. **Misleading error message**: `helpers.FailIfMultiNodes(ctx, "talm template")` hardcodes the command name in `pkg/engine/engine.go:210`. A user running `talm apply` will see an error about `talm template`, which is confusing. Consider threading the caller command name through `engine.Options`.

3. **Vestigial outer `err` check**: after the direct-patch refactor uses `if err := withApplyClient(...); err != nil { return err }`, the outer `if err != nil` at the end of the loop body only catches the template path. Consider restructuring so both branches return errors uniformly — the current state is fragile to future refactors.

4. **Dead field**: `engine.Options.Insecure` is set by `buildApplyRenderOptions` but never read anywhere in `pkg/engine/`. Safe to remove.

### Process suggestion: TDD for bug fixes

For future bug fixes, please consider writing the test first:

1. **Red**: write a test that reproduces the bug and fails on current code. This proves the bug exists and the test can detect it.
2. **Green**: apply the minimal fix that makes the test pass.
3. **Refactor**: clean up without changing behavior.

This is especially valuable for bugs like this one, where the root cause is subtle (offline vs online lookup semantics) and the fix is a one-line config change that is easy to accidentally revert. A failing-first test is the only durable guarantee that the bug stays fixed.

Happy to help with the test if useful — the `pkg/engine/render_test.go` pattern from #116 (offline rendering tests with a swappable LookupFunc) would be a good starting point.

Thanks for the quick turnaround on this one!

lexfrei (approved):
Regression test solidly reproduces the bug (offline → empty interface, online → populated interface) and exercises the real `talm.discovered.default_link_name_by_gateway` lookup pattern. All tests pass locally. Thanks for the follow-up!

The bonus action items from the previous review are tracked as separate issues so they can be picked up independently.

Issue #125, opened 2026-04-14
feat(talm): discover secondary network links alongside primary gateway link


lexfrei:
## Problem

`charts/talm/templates/_helpers.tpl` exposes `talm.discovered.default_link_name_by_gateway`, which returns the link carrying the default route. The cozystack preset (`charts/cozystack/templates/_helpers.tpl:78`) uses that helper to auto-configure a single primary interface.

On nodes with multiple NICs, every non-primary NIC is invisible to the template. A typical case is a control-plane node with:

- a primary NIC carrying the default gateway (uplink),
- one or more secondary NICs on a separate L2/VLAN segment used for node IP, shared floating IP (VIP), and cluster-internal API endpoint.

Today the secondary NIC has to be pinned by kernel name (`ens5`, `ens6`, `ens1f0`, …) in a per-node override. Kernel names are unstable — they depend on VM shape/platform, so `ens5` becomes `ens6` after a shape change and the rendered config silently breaks.

## Current state

- `talm.discovered.default_link_name_by_gateway` — returns the gateway link only.
- `talm.discovered.physical_links_info` — iterates all physical links but is documentation only (renders a comment block).
- `talm.discovered.bond_slaves` — established pattern for emitting a list of link names as JSON.
- No helper enumerates non-primary physical links.

## Proposal

Add a single, narrowly-scoped helper to `charts/talm/templates/_helpers.tpl`, following the existing `bond_slaves` shape (list + `toJson`):

```gotmpl
{{- define "talm.discovered.secondary_link_names" -}}
{{- $defaultLink := include "talm.discovered.default_link_name_by_gateway" . | trim -}}
{{- $names := list -}}
{{- range (lookup "links" "" "").items -}}
{{- if and .spec.busPath (regexMatch "^(eno|eth|enp|enx|ens)" (.metadata.id | toString)) -}}
{{- if ne .metadata.id $defaultLink -}}
{{- $names = append $names .metadata.id -}}
{{- end -}}
{{- end -}}
{{- end -}}
{{- toJson (sortAlpha $names) -}}
{{- end -}}
```

Consumed the same way as `bond_slaves`:

```gotmpl
{{- $secondary := fromJsonArray (include "talm.discovered.secondary_link_names" .) -}}
{{- if $secondary }}{{ index $secondary 0 }}{{ end }}
```

Notes:

- Output is a JSON array and is deterministic. Callers that index by position (`index $links 0`) need a stable order.
[...]

PR #127, opened 2026-04-16
fix(network): correct multi-NIC discovery and add per-link helpers

### LINKED ISSUE #58: Node default configuration results in invalid yaml through talm template
I have a worker node, whose network configuration was the default (no modifications needed, dhcp assigned, etc.). The original worker.yaml contained:

```yaml
...snip...
network: {}
...snip...
```

When attempting to migrate to using `talm`, I hit this error when running a command such as: `talm --context my_cluster -e <endpoint> -n <node id> template -t templates/worker.yaml` (Using the original and unmodified rendered templates and talm chart via `talm init`).

```shell
# talm --context my_cluster -e <endpoint> -n <node id> template -t templates/worker.yaml      
failed to render templates: decode error: yaml: line 40: could not find expected ':'
```

The debug output resulted in the problem section:

```yaml
  network:
    hostname: "MYWORKERNODE1"
    nameservers: ["<my.ga.te.way>"]
    
    # -- Discovered interfaces:
    # enp1s0:
    #   hardwareAddr:00:02:c9:de:ad:be
    #   busPath: 0000:01:00.0
    #   driver: mlx4_core
    #   vendor: Mellanox Technologies
    #   product: MT26448 [ConnectX EN 10GigE, PCIe 2.0 5GT/s])
    # enp1s0d1:
    #   hardwareAddr:00:02:c9:de:ad:bf
    #   busPath: 0000:01:00.0
    #   driver: mlx4_core
    #   vendor: Mellanox Technologies
    #   product: MT26448 [ConnectX EN 10GigE, PCIe 2.0 5GT/s])
    interfaces:
    
    []
```

The problem is the `[]` being 2 lines below.

Expected result would exclude the `interfaces` object at all, I think. Or maybe even the whole `network` object. But, it's possible my worker's configuration isn't entirely valid, either. But it does work as it currently is.

lexfrei:
## Summary

Three independent fixes/improvements in the discovery helpers and network templates, all touching the same area:

1. **`default_*_by_gateway` helpers now return a single deterministic value.** On multi-NIC nodes with several default routes (typical with DHCP), the helpers iterated every matching route and concatenated the outputs — `eth0eth1` for `default_link_name_by_gateway`, two MACs glued together for the address variant, two `busPath: ...` YAML fragments for the selector. Downstream templates used the joined string as the interface name, so Talos either rejected the config or applied it to the wrong NIC. They now filter by `table=main` and break after the first match. Also fixes a long-standing copy-paste bug in `default_link_bus_by_gateway`, which was returning the link's MAC address instead of its `busPath`.

2. **Empty interface emission removed from cozystack/generic templates.** When discovery returned no default route (offline render, isolated node, custom networking), the templates emitted Talos documents with empty names: `LinkConfig`/`BondConfig`/`VLANConfig`/`Layer2VIPConfig` with empty `name:` (multidoc), or `interfaces:` followed by `- interface:` with an empty value (legacy). Talos v1.12 rejects the multidoc form; the legacy form produced invalid YAML in some environments. The per-interface documents are now skipped when no default link is discovered.

3. **New per-link discovery helpers for multi-NIC templates.** The existing `default_*_by_gateway` family exposes only the primary link. Templates for nodes with secondary NICs (storage uplink on a control-plane, second public link, etc.) had no way to enumerate every NIC or read addresses/gateway/MAC by name. Adds `physical_link_names`, `configurable_link_names`, `addresses_by_link`, `gateway_by_link` (IPv4 only by convention), `routes_by_link`, `mac_by_link`, `bus_by_link`, `link_selector_by_name`. Default cozystack/generic templates are unchanged; the new helpers are building blocks for user templates.

## Verification

- New unit tests cover all three changes: `TestDefaultLinkByGatewayHelpers_MultiNIC`, `TestNetworkMultidoc_NoDiscovery` / `TestNetworkLegacy_NoDiscovery`, `TestSecondaryNicHelpers`. The multi-NIC test fixture deliberately mixes routes across `table=main` and a private table to exercise both the table filter and the early break; the secondary-NIC fixture orders an IPv6 default route ahead of the IPv4 one to catch a missing family filter; the route fixture includes both an entry with absent fields and one with `priority: 0`.
- Existing offline tests that previously asserted broken-empty `LinkConfig` output now use `simpleNicLookup` so they continue to validate the populated render path. `TestMultiDocCozystack_BondTopology` and `_VlanOnBondTopology` are byte-for-byte identical and still pass.
- `golangci-lint run` is clean.

## Closes

- Closes #58
- Closes #108
- Refs #125 — adds the helpers; cozystack default templates intentionally unchanged

## Release Notes

* **New Features**
  * Added multi-NIC network helper utilities for enhanced network configuration discovery and management.

* **Bug Fixes**
  * Improved network discovery to accurately select the correct network interface when multiple default routes exist.
  * Enhanced network interface configuration rendering to gracefully handle scenarios where network discovery provides no data.

* **Tests**
  * Added comprehensive regression tests for multi-NIC and network discovery scenarios to ensure robust network configuration handling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 talm
 dist/
+.claude/worktrees/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 talm
 dist/
-.claude/worktrees/
--- a/charts/talm/templates/_helpers.tpl
+++ b/charts/talm/templates/_helpers.tpl
@@ -126,7 +126,7 @@
 {{- define "talm.discovered.default_link_bus_by_gateway" }}
 {{- range (lookup "routes" "" "").items }}
 {{- if and (eq .spec.dst "") (not (eq .spec.gateway "")) (eq .spec.table "main") }}
-{{- (lookup "links" "" .spec.outLinkName).spec.hardwareAddr }}
+{{- (lookup "links" "" .spec.outLinkName).spec.busPath }}
 {{- break }}
 {{- end }}
 {{- end }}
@@ -282,8 +282,10 @@ vlans:
   remain wrappers that resolve the primary link and call into these.
 */ -}}
 
-{{- /* JSON list of physical link names (raw NICs only — not bond/vlan masters). */ -}}
-{{- define "talm.discovered.physical_links" -}}
+{{- /* JSON list of physical link names (raw NICs only — not bond/vlan masters).
+       Renamed from `physical_links` to avoid collision with the older
+       `physical_links_info` helper, which renders YAML comments. */ -}}
+{{- define "talm.discovered.physical_link_names" -}}
 {{- $names := list -}}
 {{- range (lookup "links" "" "").items -}}
 {{- if and .spec.busPath (regexMatch "^(eno|eth|enp|enx|ens)" (.metadata.id | toString)) -}}
@@ -295,7 +297,7 @@ vlans:
 
 {{- /* JSON list of every link a user template can configure: physical NICs
        plus bond / vlan / bridge top-level links. */ -}}
-{{- define "talm.discovered.configurable_links" -}}
+{{- define "talm.discovered.configurable_link_names" -}}
 {{- $names := list -}}
 {{- range (lookup "links" "" "").items -}}
 {{- $isPhysical := and .spec.busPath (regexMatch "^(eno|eth|enp|enx|ens)" (.metadata.id | toString)) -}}
@@ -308,15 +310,17 @@ vlans:
 {{- end -}}
 
 {{- /* JSON list of CIDR addresses configured on the given link, excluding
-       kernel-managed scopes (host loopback, link-local, nowhere). Both IPv4
-       and IPv6 globally-scoped addresses are returned; the caller is
-       responsible for filtering by family or stripping VIPs if needed. */ -}}
+       kernel-managed scopes (host loopback, link-local, nowhere) and
+       addresses with no scope set at all. Both IPv4 and IPv6 globally-scoped
+       addresses are returned; the caller is responsible for filtering by
+       family or stripping VIPs if needed. */ -}}
 {{- define "talm.discovered.addresses_by_link" -}}
 {{- $linkName := . -}}
 {{- $addresses := list -}}
 {{- range (lookup "addresses" "" "").items -}}
+{{- $hasScope := and .spec.scope (ne (.spec.scope | toString) "") -}}
 {{- $skip := has (.spec.scope | toString) (list "host" "link" "nowhere") -}}
-{{- if and (eq .spec.linkName $linkName) (not $skip) -}}
+{{- if and (eq .spec.linkName $linkName) $hasScope (not $skip) -}}
 {{- $addresses = append $addresses .spec.address -}}
 {{- end -}}
 {{- end -}}
@@ -339,37 +343,43 @@ vlans:
 
 {{- /* JSON list of non-default routes on the given link. Each entry is a flat
        map {dst, gateway, family, table, priority} so consumers can
-       fromJsonArray + range + dig. Missing route fields are emitted as empty
-       strings so consumers never see the literal "<nil>". */ -}}
+       fromJsonArray + range + dig. Absent fields are emitted as empty
+       strings; present fields including the integer 0 (e.g. priority: 0)
+       round-trip through toString. */ -}}
 {{- define "talm.discovered.routes_by_link" -}}
 {{- $linkName := . -}}
 {{- $routes := list -}}
 {{- range (lookup "routes" "" "").items -}}
 {{- if and (eq .spec.outLinkName $linkName) (not (eq .spec.dst "")) -}}
-{{- $entry := dict
-    "dst" .spec.dst
-    "gateway" (default "" .spec.gateway | toString)
-    "family" (default "" .spec.family | toString)
-    "table" (default "" .spec.table | toString)
-    "priority" (default "" .spec.priority | toString) -}}
+{{- $gateway := "" -}}
+{{- if not (kindIs "invalid" .spec.gateway) -}}{{- $gateway = printf "%v" .spec.gateway -}}{{- end -}}
+{{- $family := "" -}}
+{{- if not (kindIs "invalid" .spec.family) -}}{{- $family = printf "%v" .spec.family -}}{{- end -}}
+{{- $table := "" -}}
+{{- if not (kindIs "invalid" .spec.table) -}}{{- $table = printf "%v" .spec.table -}}{{- end -}}
+{{- $priority := "" -}}
+{{- if not (kindIs "invalid" .spec.priority) -}}{{- $priority = printf "%v" .spec.priority -}}{{- end -}}
+{{- $entry := dict "dst" .spec.dst "gateway" $gateway "family" $family "table" $table "priority" $priority -}}
 {{- $routes = append $routes $entry -}}
 {{- end -}}
 {{- end -}}
 {{- toJson $routes -}}
 {{- end -}}
 
-{{- /* Scalar MAC address for the given link, or empty. */ -}}
+{{- /* Scalar MAC address for the given link, or empty if the link or its
+       spec is missing (test mocks may return {} for unknown ids). */ -}}
 {{- define "talm.discovered.mac_by_link" -}}
 {{- $link := lookup "links" "" . -}}
-{{- if $link -}}
+{{- if and $link $link.spec -}}
 {{- $link.spec.hardwareAddr | toString -}}
 {{- end -}}
 {{- end -}}
 
-{{- /* Scalar PCI / bus path for the given link, or empty. */ -}}
+{{- /* Scalar PCI / bus path for the given link, or empty if the link or its
+       spec is missing. */ -}}
 {{- define "talm.discovered.bus_by_link" -}}
 {{- $link := lookup "links" "" . -}}
-{{- if $link -}}
+{{- if and $link $link.spec -}}
 {{- $link.spec.busPath | toString -}}
 {{- end -}}
 {{- end -}}
--- a/charts/talm/templates/_helpers.tpl
+++ b/charts/talm/templates/_helpers.tpl
@@ -307,26 +307,30 @@ vlans:
 {{- toJson $names -}}
 {{- end -}}
 
-{{- /* JSON list of CIDR addresses configured on the given link (any family),
-       excluding host-scoped addresses. Caller is responsible for filtering
-       VIPs or family if needed. */ -}}
+{{- /* JSON list of CIDR addresses configured on the given link, excluding
+       kernel-managed scopes (host loopback, link-local, nowhere). Both IPv4
+       and IPv6 globally-scoped addresses are returned; the caller is
+       responsible for filtering by family or stripping VIPs if needed. */ -}}
 {{- define "talm.discovered.addresses_by_link" -}}
 {{- $linkName := . -}}
 {{- $addresses := list -}}
 {{- range (lookup "addresses" "" "").items -}}
-{{- if and (eq .spec.linkName $linkName) (not (eq .spec.scope "host")) -}}
+{{- $skip := has (.spec.scope | toString) (list "host" "link" "nowhere") -}}
+{{- if and (eq .spec.linkName $linkName) (not $skip) -}}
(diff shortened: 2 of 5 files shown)


charts/talm/templates/_helpers.tpl
```diff
@@ -106,31 +107,34 @@
...
+{{- break }}
 {{- end }}
 {{- end }}
 {{- end }}
 
 {{- define "talm.discovered.default_link_bus_by_gateway" }}
 {{- range (lookup "routes" "" "").items }}
-{{- if and (eq .spec.dst "") (not (eq .spec.gateway "")) }}
+{{- if and (eq .spec.dst "") (not (eq .spec.gateway "")) (eq .spec.table "main") }}
 {{- (lookup "links" "" .spec.outLinkName).spec.hardwareAddr }}
```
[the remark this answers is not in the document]
lexfrei: Fixed in b3729ad. Same diagnosis — the helper had been returning `.spec.hardwareAddr` since 2024-05 instead of `.spec.busPath`. Changed exactly as suggested. Coverage added in `TestDefaultLinkByGatewayHelpers_MultiNIC`, which asserts `bus=pci-0000:00:1f.0` and explicitly rejects the MAC.

charts/talm/templates/_helpers.tpl
```diff
@@ -266,3 +270,106 @@ vlans:
...
+
+{{- /* JSON list of non-default routes on the given link. Each entry is a flat
+       map {dst, gateway, family, table, priority} so consumers can
+       fromJsonArray + range + dig. */ -}}
+{{- define "talm.discovered.routes_by_link" -}}
+{{- $linkName := . -}}
+{{- $routes := list -}}
+{{- range (lookup "routes" "" "").items -}}
+{{- if and (eq .spec.outLinkName $linkName) (not (eq .spec.dst "")) -}}
+{{- $entry := dict "dst" .spec.dst "gateway" (.spec.gateway | toString) "family" (.spec.family | toString) "table" (.spec.table | toString) "priority" (.spec.priority | toString) -}}
```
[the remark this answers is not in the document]
lexfrei: Addressed in b3729ad with a slightly different choice. Each optional field is now wrapped in `kindIs "invalid"` and emitted via `printf "%v"`, so absent fields become `""` (never `"<no value>"` or `"<nil>"`) and present values — including integer `0` (e.g. `priority: 0`) — round-trip correctly. Values are kept as strings rather than letting `toJson` emit `null` for missing fields, because consumer templates iterate via `fromJsonArray` + `dig` and benefit from a uniform string contract. `TestSecondaryNicHelpers` covers both the absent-fields and `priority: 0` paths and asserts no `<nil>`/`<no value>` leaks.

charts/talm/templates/_helpers.tpl
```diff
@@ -266,3 +270,106 @@ vlans:
...
+{{- end -}}
+{{- end -}}
+{{- toJson $routes -}}
+{{- end -}}
+
+{{- /* Scalar MAC address for the given link, or empty. */ -}}
+{{- define "talm.discovered.mac_by_link" -}}
+{{- $link := lookup "links" "" . -}}
+{{- if $link -}}
+{{- $link.spec.hardwareAddr | toString -}}
```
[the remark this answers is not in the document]
lexfrei: Fixed in b3729ad. Both helpers now check `and $link $link.spec`, so an unknown link name yields an empty string in either regime (real Helm `lookup` returns `nil`, test mocks may return `{}`). `TestSecondaryNicHelpers` asserts `mac_unknown=` and `bus_unknown=` against an unknown link to pin the contract.

.gitignore
```diff
@@ -1,2 +1,3 @@
 talm
 dist/
+.claude/worktrees/
```
myasnikovdaniil: **contributor-specific entry**

```text
.claude/worktrees/
```

This is Claude-Code-specific scratch state and shouldn't live in the project's `.gitignore` — it leaks one contributor's tooling onto everyone. Recommend putting it in your personal `~/.config/git/ignore` (or `core.excludesFile`) instead and dropping the commit `06581ec` from this PR.

```text
talm
dist/
```

charts/talm/templates/_helpers.tpl
```diff
@@ -266,3 +270,126 @@ vlans:
...
+{{- end -}}
+
+{{- /* Scalar PCI / bus path for the given link, or empty if the link or its
+       spec is missing. */ -}}
+{{- define "talm.discovered.bus_by_link" -}}
+{{- $link := lookup "links" "" . -}}
+{{- if and $link $link.spec -}}
+{{- $link.spec.busPath | toString -}}
+{{- end -}}
+{{- end -}}
```
myasnikovdaniil, line 391: **`bus_by_link` / `mac_by_link` can emit `<nil>` for virtual links**

```yaml
{{- define "talm.discovered.bus_by_link" -}}
{{- $link := lookup "links" "" . -}}
{{- if and $link $link.spec -}}
{{- $link.spec.busPath | toString -}}
{{- end -}}
{{- end -}}
```

When the link exists with a `spec` but the specific field is absent — concretely, `bus_by_link "bond0"` on the secondary-NIC fixture, since virtual links don't have a `busPath` — `$link.spec.busPath` is nil and `nil | toString` renders as `<nil>`. The doc string promises "empty if the link or its spec is missing", but a present spec with a missing field falls through. The global `assertNotContains "<nil>"` in `TestSecondaryNicHelpers` doesn't exercise this exact call.

Suggested tighten:

```yaml
{{- define "talm.discovered.bus_by_link" -}}
{{- $link := lookup "links" "" . -}}
{{- if and $link $link.spec $link.spec.busPath -}}
{{- $link.spec.busPath | toString -}}
{{- end -}}
{{- end -}}
```

Same shape for `mac_by_link` (guard on `$link.spec.hardwareAddr`).

myasnikovdaniil (commented):
Below is the draft review. Tests pass locally on the PR HEAD (combined with `00358fd` from origin/main, which the PR is rebased on); flagged a couple of minor items but nothing blocking.

---

# Review of PR #127 — `fix(network): correct multi-NIC discovery and add per-link helpers`

## Summary

Three independent and well-motivated changes in the discovery layer:

1. The `default_*_by_gateway` family is fixed to return a single deterministic value — adds the missing `table=main` filter and breaks after the first match, ending the `eth0eth1` concatenation on multi-NIC DHCP nodes. Also fixes a real copy-paste bug in `default_link_bus_by_gateway` that returned the MAC instead of the busPath.
2. Empty `LinkConfig`/`BondConfig`/`VLANConfig`/`Layer2VIPConfig` documents (multidoc) and bare `- interface:` blocks (legacy) are no longer emitted when discovery returns nothing — Talos v1.12 was rejecting these.
3. New per-link helpers (`physical_link_names`, `addresses_by_link`, `gateway_by_link`, `routes_by_link`, `mac_by_link`, `bus_by_link`, `link_selector_by_name`, `configurable_link_names`) so user templates can configure secondary NICs without monkey-patching the existing primary-link wrappers.

The bugfixes are real and the test fixtures are thoughtfully constructed (mixed routing tables, IPv6 ordered before IPv4, `priority: 0` and missing-field round-trip). All tests pass locally; the existing offline `TestMultiDocCozystack_*` tests are correctly migrated to use a populated `simpleNicLookup` so they stop relying on the broken-empty render path.

A couple of minor observations below — none are blockers.

## Recommendation

**Comment** (lean approve). The three core fixes are correct and well-tested; the minor items above are tightening suggestions, not blockers. I'd at least drop the `.claude/worktrees/` `.gitignore` entry before merging.

charts/talm/templates/_helpers.tpl
```diff
@@ -266,3 +270,126 @@ vlans:
...
+{{- $table := "" -}}
+{{- if not (kindIs "invalid" .spec.table) -}}{{- $table = printf "%v" .spec.table -}}{{- end -}}
+{{- $priority := "" -}}
+{{- if not (kindIs "invalid" .spec.priority) -}}{{- $priority = printf "%v" .spec.priority -}}{{- end -}}
+{{- $entry := dict "dst" .spec.dst "gateway" $gateway "family" $family "table" $table "priority" $priority -}}
+{{- $routes = append $routes $entry -}}
+{{- end -}}
+{{- end -}}
+{{- toJson $routes -}}
+{{- end -}}
```
myasnikovdaniil, line 369: **`routes_by_link` returns `priority` as a string**

```yaml
{{- $priority := "" -}}
{{- if not (kindIs "invalid" .spec.priority) -}}{{- $priority = printf "%v" .spec.priority -}}{{- end -}}
{{- $entry := dict "dst" .spec.dst "gateway" $gateway "family" $family "table" $table "priority" $priority -}}
```

The result is `"priority":"0"` / `"priority":"100"` — string-typed in the JSON. Functional and documented in the helper comment, but downstream callers that want metric/priority ordering will have to `atoi` the value. If you'd rather keep it native:

```yaml
{{- $entry := dict "dst" .spec.dst "gateway" $gateway "family" $family "table" $table -}}
{{- if not (kindIs "invalid" .spec.priority) -}}
{{- $entry = merge $entry (dict "priority" .spec.priority) -}}
{{- end -}}
```

Not a blocker; flagging because `priority` is the one field where the lossy stringification is most likely to bite a future consumer.

charts/talm/templates/_helpers.tpl
```diff
@@ -266,3 +270,126 @@ vlans:
...
+{{- break -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- /* JSON list of non-default routes on the given link. Each entry is a flat
+       map {dst, gateway, family, table, priority} so consumers can
+       fromJsonArray + range + dig. Absent fields are emitted as empty
+       strings; present fields including the integer 0 (e.g. priority: 0)
+       round-trip through toString. */ -}}
```
myasnikovdaniil: **`routes_by_link` doc could mention the table behaviour**

The helper doc comment lists the returned fields but doesn't say it returns routes from **all** tables (not just `main`). Since `gateway_by_link` filters to `table=main` but `routes_by_link` doesn't, calling out the asymmetry would save a future reader a grep. One-line suggestion:

```yaml
{{- /* JSON list of non-default routes on the given link, across all routing
       tables (the `table` field is included so callers can filter). Each
       entry is a flat map {dst, gateway, family, table, priority} ...
*/ -}}
```

.gitignore
lexfrei: Removed in ab48e7f and folded the entry into my personal `~/.config/git/ignore` per your suggestion, so future runs don't re-surface the path in `git status` and re-tempt me to commit it.

charts/talm/templates/_helpers.tpl
lexfrei: Tightened both helpers in 28f37e0 — the `if` guard now extends to the specific field (`$link.spec.busPath` / `$link.spec.hardwareAddr`), so a present-spec / absent-field link returns empty rather than dereferencing through `nil | toString`. The `bond0` fixture in `secondaryNicLookup` already had `hardwareAddr` but no `busPath`; added explicit `bus_bond0` / `mac_bond0` template inclusions to `TestSecondaryNicHelpers` so the regression is pinned directly rather than relying on the global `assertNotContains "<nil>"` sweep. Updated the doc comments to spell out the present-spec / absent-field case.

charts/talm/templates/_helpers.tpl
lexfrei: Declining this one. The helper's invariant is "all fields are present strings; absent fields are empty strings", per the existing doc and pinned by the `assertContains '"priority":"0"'` assertion. Switching `priority` alone to native int introduces a heterogeneous shape — "string for `dst`, int for `priority`, empty string for absent fields" — which forces every consumer to add a type/presence check. The single-field `atoi` cost on the consumer side is real but bounded.

Happy to revisit if a downstream consumer surfaces concrete metric-ordering pain — at that point the better answer is probably a sibling helper that emits native types throughout, leaving this one as the string-uniform version.

charts/talm/templates/_helpers.tpl
lexfrei: Updated in 28f37e0 — the doc now calls out the all-tables behaviour and the asymmetry with `gateway_by_link` (which narrows to `table=main`).

kvaps (approved):
LGTM. Multi-NIC fixes are well-engineered with comprehensive regression tests covering all edge cases (table filter, family filter, missing fields, priority 0, scope filtering).

PR #128, opened 2026-04-16
fix(commands): apply batch — multi-node, node-file patch merge, error handling

### LINKED ISSUE #120: talm apply with templates rejects multi-node modelines in online rendering
## Context

PR #119 moved template rendering in `talm apply` from offline to online mode (rendering inside the Talos client callback so `lookup()` resolves real discovery data). This fixes the reported Talos v1.12 validation error, but introduces a behavioral change for multi-node modelines.

## Problem

`engine.Render` calls `helpers.FailIfMultiNodes` (`pkg/engine/engine.go:210`) whenever `Offline=false`. Previously, `talm apply` rendered offline, so this check never fired, and the subsequent `ApplyConfiguration` gRPC call handled multi-node fan-out at the wire level.

After #119, users who:
- put multiple IPs in a single modeline `nodes=[ip1, ip2]`, or
- pass `--nodes ip1,ip2` on the command line

will hit `FailIfMultiNodes` before any rendering happens, in non-insecure mode.

## Unaffected paths

- The documented single-node-per-file flow (`-f node1.yaml -f node2.yaml`) still works because each file modeline sets exactly one node.
- Maintenance mode (`-i`) still works because `WithClientMaintenance` does not inject `nodes` metadata into the gRPC context.

## Proposed fix

Decide intent and apply one of:
- Document the single-node-per-file convention explicitly in README.
- Render once per node in the client callback when multiple nodes are present.
- Reject multi-node modelines up front with a clearer error message explaining the constraint.

## References

- Introduced in #119
- `pkg/engine/engine.go:209-213`
- `pkg/commands/apply.go:120` (withApplyClient callback)

### LINKED ISSUE #121: FailIfMultiNodes hardcodes "talm template" in error message, confusing apply callers
## Problem

`pkg/engine/engine.go:210` calls:

```go
helpers.FailIfMultiNodes(ctx, "talm template")
```

The string `"talm template"` is used in the error message: `command "talm template" is not supported with multiple nodes`. This was accurate when `engine.Render` was only called from the `template` subcommand, but after #119 it is also called from `apply`. A user running `talm apply` who hits this check will see an error mentioning `talm template`, which is confusing.

## Proposed fix

Thread the caller command name through `engine.Options`:

```go
type Options struct {
    // ...
    CommandName string // e.g., "talm apply" or "talm template"
}
```

Then use `opts.CommandName` in the `FailIfMultiNodes` call, with a sensible default.

## References

- `pkg/engine/engine.go:209-213`
- Surfaced by #119

### LINKED ISSUE #122: apply.go has vestigial outer err check after direct-patch refactor
## Problem

After #119, `pkg/commands/apply.go` has two code paths in the per-file loop:

1. Template rendering path: assigns `err = withApplyClient(...)` at line 120
2. Direct patch path: uses `if err := withApplyClient(...); err != nil { return err }` at line 156 (returns inside the inner block)

The outer `if err != nil { return err }` at line ~176 only catches errors from the template path. Any future refactor that removes an early return inside the direct-patch block could silently swallow errors — exactly the bug the reviewers flagged on the original PR (commit b34781e fixed the shadowing).

## Proposed fix

Restructure so both branches use the same error handling idiom. Options:
- Both branches use `err := withApplyClient(...); if err != nil { return err }` with local scope.
- Or hoist shared post-render logic into a helper, so the main loop body has a single assignment → check point.

Remove the outer `if err != nil` check once both branches handle errors locally.

## References

- `pkg/commands/apply.go:120, 156, ~176`
- Introduced in #119

### LINKED ISSUE #123: engine.Options.Insecure is set but never read
## Problem

`pkg/engine/engine.go` declares `Insecure bool` on `Options`, and `buildApplyRenderOptions` in `pkg/commands/apply.go` assigns `applyCmdFlags.insecure` to it. However, `engine.Render` and all helpers in `pkg/engine/` never read `opts.Insecure`.

This was extended by #119 which added the assignment (and a regression test for the assignment), but the field still has no effect on behavior.

## Proposed fix

Either:
- Remove the field if it is truly unused.
- Wire it up if it was meant to influence rendering (e.g., suppress `FailIfMultiNodes` for the maintenance-mode path, which conceptually supports multi-node).

## References

- `pkg/engine/engine.go:55`
- `pkg/commands/apply.go:216` (in `buildApplyRenderOptions`)
- `pkg/commands/apply_test.go` — regression test asserts the pass-through

### LINKED ISSUE #126: bug(apply): node file content silently discarded when modeline contains templates=
## Summary

When `talm apply -f nodes/node0.yaml` processes a file whose modeline contains `templates=["templates/controlplane.yaml"]`, the non-modeline content of the node file (hostname, interfaces, etcd extraArgs, apiServer certSANs, time servers, etc.) is silently discarded. Only the rendered template output is sent to the node.

This means per-node customizations defined in node files have no effect when templates are specified in the modeline — which is the primary intended workflow.

## Environment

- talm v0.23.1
- Talos v1.12.6
- 3-node control plane, two NICs per node (primary + VLAN via deviceSelector)

## Reproduction

Node file (`nodes/node0.yaml`):

```yaml
# talm: nodes=["10.0.0.1"], endpoints=["10.0.0.1"], templates=["templates/controlplane.yaml"]
machine:
  network:
    hostname: node0
    interfaces:
      - interface: ens3
        addresses: ["10.0.0.1/24"]
        routes:
          - network: 0.0.0.0/0
            gateway: 10.0.0.254
      - deviceSelector:
          hardwareAddr: "02:00:17:02:55:aa"
        addresses: ["10.0.100.11/24"]
        vip:
          ip: 10.0.100.10
```

Template auto-discovers the default gateway interface and renders:

```yaml
machine:
  network:
    hostname: "talos-25957"   # auto-generated
    interfaces:
      - interface: ens3       # only primary NIC
        addresses: ["10.0.0.1/24"]
        routes:
          - network: 0.0.0.0/0
            gateway: 10.0.0.254
```

After `talm apply --insecure -f nodes/node0.yaml`, the applied config on the node has:

- `hostname: talos-25957` (template value, not `node0` from the node file)
- Only `ens3` interface (deviceSelector interface from the node file is missing)
- No VIP on the correct interface

Verified with:

```bash
talosctl --insecure -e 10.0.0.1 -n 10.0.0.1 get hostname
talosctl --insecure -e 10.0.0.1 -n 10.0.0.1 get addresses
```

## Root cause

In `pkg/commands/apply.go:107-141`, when `modelineTemplates` is non-empty (template rendering path):

1. `processModelineAndUpdateGlobals(configFile, ...)` (root.go:222) extracts only the modeline (nodes, endpoints, templates) from the node file. The rest of the file content is not captured.

2. `buildApplyRenderOptions(modelineTemplates, ...)` constructs `engine.Options` using only the template paths. The node file is not included as a patch source.

3. `engine.Render(ctx, c, opts)` renders the template and returns the full config.

4. The result is sent directly to `c.ApplyConfiguration()` — the node file content is never merged.

The variable `configFile` after line 108 is only used in a `Printf` log on line 121. Its content is never read or applied as a patch.

Compare to the direct patch path (lines 142-146, when modeline has no `templates=`):

```go
patches := []string{"@" + configFile}  // file IS used as a patch
configBundle, machineType, err := engine.FullConfigProcess(ctx, opts, patches)
```

Here the file content is correctly used as a config patch.

## Impact

- All per-node customizations are lost: hostname, secondary interfaces, VIP placement, etcd tuning, per-node certSANs, time servers, etc.
- On multi-NIC setups, etcd cannot start because the interface carrying the advertised subnet IP is missing from the applied config (`Waiting for etcd spec` indefinitely)
- Bootstrap hangs forever

## Expected behavior

The template rendering path should apply the node file as a strategic merge patch on top of the rendered template output before sending the config to the node.

## Workaround

Patch the applied config after the fact:

```bash
talosctl patch machineconfig --endpoints IP --nodes IP --talosconfig talosconfig \
  --patch '[
    {"op": "replace", "path": "/machine/network/hostname", "value": "node0"},
    {"op": "replace", "path": "/machine/network/interfaces", "value": [
      {"interface": "ens3", "addresses": ["10.0.0.1/24"], "routes": [{"network": "0.0.0.0/0", "gateway": "10.0.0.254"}]},
      {"deviceSelector": {"hardwareAddr": "02:00:17:02:55:aa"}, "addresses": ["10.0.100.11/24"], "vip": {"ip": "10.0.100.10"}}
    ]}
  ]'
```

Note: the piped workaround from #113 (`talm template -f ... | talm apply -f -`) does not help here — `talm template` also only outputs the rendered template without merging the node file content.

## Introduced in

PR #113 added the template rendering path in `apply` but did not include a step to merge the node file content as a patch on top of the rendered template.

PR #119 moved rendering from offline to online but did not address the missing merge.

lexfrei:
## Summary

Five issues clustered in `pkg/commands/apply.go` and `pkg/engine/engine.go`. All were either introduced or surfaced by PR #119 (online template rendering in apply). Bundled because they share the same callback structure and overlap heavily.

1. **Per-node template rendering in `talm apply`.** PR #119 moved rendering inside the Talos client callback so `lookup()` resolves real discovery data, but the callback batched all `--nodes A,B,C` (or modeline `nodes=[A,B,C]`) into a single gRPC context. `engine.Render` then hit `helpers.FailIfMultiNodes` and bailed out before any rendering happened. Render once per node instead, each iteration carrying a single-node `client.WithNode` context so the guard passes and discovery resolves per node. In insecure (maintenance) mode each iteration opens a fresh single-endpoint maintenance client because `WithClientMaintenance` would otherwise dial every endpoint at once and gRPC would round-robin `ApplyConfiguration` across them — most nodes would never receive the config.

2. **Node file is now applied as a patch over the rendered template.** When a modeline contained `templates=[...]`, `talm apply -f node.yaml` rendered the listed templates and discarded the rest of the node file. Per-node hostname, secondary interfaces with `deviceSelector`, VIP placement, etcd extra args — all silently lost. New `engine.MergeFileAsPatch` overlays the body using the Talos config-patcher; modeline-only and effectively-empty files short-circuit before reaching the patcher (otherwise the patcher routes empty input through JSON6902 and rejects any multi-document rendered config — the v1.12+ default output format). `talm template` intentionally does not call this overlay because the patcher round-trips through its config loader, dropping the modeline and YAML comments downstream commands rely on; the README documents the divergence.

3. **Multi-node error message references the actual subcommand.** `helpers.FailIfMultiNodes` was hardcoded to `"talm template"`. After #119, users running `talm apply` saw an error talking about `talm template`. Threaded `CommandName` through `engine.Options`; defaults to `"talm"` when empty.

4. **Symmetric error handling in the apply per-file loop.** The template-render branch assigned `err = withApplyClient(...)` and relied on an outer `if err != nil`; the direct-patch branch already used local-scope `if err := ...; err != nil { return err }`. Two idioms, one outer check that only caught one branch — the same shadowing class that already bit the project once. Both branches now use the local-scope return; the outer check is gone.

5. **Removed dead `engine.Options.Insecure`.** The field has been declared since the file was first added (May 2024) and was never read by `Render` or any helper. The runtime guarantee — `--insecure` (maintenance mode) bypasses `FailIfMultiNodes` — comes from `WithClientMaintenance` not injecting nodes into the gRPC context, not from anything in the engine.

## Verification

- `TestRenderFailIfMultiNodes_UsesCommandName` — three sub-tests asserting the `CommandName` value surfaces in the error and the empty-string fallback.
- `TestMergeFileAsPatch` — overlay (single-doc + non-empty patch), modeline-only byte-identity, multi-doc + modeline-only (the regression vector), multi-doc + non-empty strategic-merge patch (the realistic v1.12+ scenario), empty file, comments-only file.
- `TestApplyTemplatesPerNode_*` — per-node single-node ctx; never-batches; empty-nodes errors loudly; maintenance mode opens a fresh client per iteration; auth mode writes the single-target `node` metadata key.
- `TestOpenClientPerNodeMaintenance_NarrowsAndRestoresGlobalNodes` — drives the production opener with an injected `maintenanceClientFunc` fake that snapshots `GlobalArgs.Nodes` at the moment `WithClientMaintenance` reads it.
- `TestTemplateAndApplyDiverge_NodeBodyOverlayLimitation` — pins the intentional template/apply divergence so a future "make template match apply" change is caught.
- `TestBuildApplyRenderOptions` — pins that the apply caller actually sets `CommandName` (independent of the engine-side coverage).
- `golangci-lint run ./...` is clean.

## Closes

- Closes #120
- Closes #121
- Closes #122
- Closes #123
- Closes #126

## Release Notes

* **New Features**
  * Node YAML files can now include a modeline "overlay" body that applies as a strategic merge patch during `talm apply`, preserving per-node configurations even when templates generate conflicting values.
  * Use `talm apply --dry-run` to preview exact bytes sent to nodes.

* **Documentation**
  * Added clarification on overlay behavior differences between `talm apply` and `talm template` commands.


pkg/commands/apply.go
```diff
@@ -113,32 +113,40 @@ func apply(args []string) error {
...
+			}
+
+			if applyCmdFlags.insecure {
+				openClient := openClientPerNodeMaintenance(applyCmdFlags.certFingerprints, WithClientMaintenance)
+				if err := applyTemplatesPerNode(opts, configFile, nodes, openClient, engine.Render, applyClosure); err != nil {
+					return err
+				}
+			} else {
+				if err := withApplyClientBare(func(parentCtx context.Context, c *client.Client) error {
+					openClient := openClientPerNodeAuth(parentCtx, c)
```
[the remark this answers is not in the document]
lexfrei: Added the same fallback in 5a16ad33. Extracted `resolveAuthTemplateNodes` so the precedence (CLI/modeline first, talosconfig context second) is unit-testable and symmetric with what `wrapWithNodeContext` already did for the direct-patch branch. Insecure (maintenance) mode still uses only `GlobalArgs.Nodes` — the maintenance client talks to node IPs directly and carries no talosconfig context to consult.

pkg/engine/engine.go
```diff
@@ -202,6 +204,68 @@ func SerializeConfiguration(configBundle *bundle.Bundle, machineType machine.Typ
...
+		return nil, fmt.Errorf("reading patch %s: %w", patchFile, err)
+	}
+	if isEffectivelyEmptyYAML(patchBytes) {
+		return rendered, nil
+	}
+	patch, err := configpatcher.LoadPatch(patchBytes)
+	if err != nil {
+		return nil, fmt.Errorf("loading patch from %s: %w", patchFile, err)
+	}
+	out, err := configpatcher.Apply(configpatcher.WithBytes(rendered), []configpatcher.Patch{patch})
```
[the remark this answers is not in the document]
lexfrei: Checked the Talos source — `configpatcher.LoadPatch` already routes through `configloader.NewFromBytes` first, which handles multi-document Talos configs as a strategic-merge patch. A multi-doc node body therefore already works today; `configpatcher.LoadPatches([]string{string(patchBytes)})` with a single-element slice would resolve to the same `LoadPatch` call and add no capability. `LoadPatches` only splits its input when the caller supplies multiple strings (typically `@file` references), not by splitting a single string on `---`. Keeping `LoadPatch` here.

pkg/commands/apply.go
```diff
@@ -188,23 +193,131 @@ func apply(args []string) error {
...
+func applyTemplatesPerNode(
+	opts engine.Options,
+	configFile string,
+	nodes []string,
+	openClient openClientFunc,
+	render renderFunc,
+	apply applyFunc,
+) error {
+	if len(nodes) == 0 {
+		return fmt.Errorf("no nodes specified for template-rendering apply")
```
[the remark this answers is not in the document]
lexfrei: Reworded in 599079d3 to name the three concrete ways to set nodes (`--nodes`, modeline, talosconfig context) and strengthened `TestApplyTemplatesPerNode_NoNodesIsAnError` to assert the guidance stays put through future rewords. With the talosconfig-context fallback from 5a16ad33 this error now only fires when all three sources are empty.

pkg/engine/engine.go
```diff
@@ -202,6 +204,68 @@ func SerializeConfiguration(configBundle *bundle.Bundle, machineType machine.Typ
...
+// JSON6902 and reject any multi-document rendered config (the v1.12+ output
+// format) outright.
+//
+// Note that for non-empty patches the patcher round-trips rendered through
+// its config loader, normalising YAML formatting and dropping comments.
+// This is acceptable for the apply path (the result goes straight to
+// ApplyConfiguration) but unsuitable for human-facing output such as
+// `talm template` — which is why the template subcommand does not call
+// this helper.
+func MergeFileAsPatch(rendered []byte, patchFile string) ([]byte, error) {
```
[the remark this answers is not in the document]
lexfrei: After the multi-node + non-empty body guard in 66a94892, any non-empty patch is guaranteed to target exactly one node — so `MergeFileAsPatch` reads and parses the file once per apply. For a single-node call the read/parse is ≤1 ms and comfortably dominated by template render and gRPC apply; a refactor to split file I/O from the merge would pay for itself only on large node-body files applied to hundreds of nodes, which the new guard forbids. Happy to revisit if a use case emerges that the guard does not cover.

pkg/commands/apply.go
```diff
@@ -188,23 +193,131 @@ func apply(args []string) error {
...
+	}
+	for _, node := range nodes {
+		if err := openClient(node, func(ctx context.Context, c *client.Client) error {
+			return renderMergeAndApply(ctx, c, opts, configFile, render, apply)
+		}); err != nil {
```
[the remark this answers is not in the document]
lexfrei: Fixed in 66a94892. `applyTemplatesPerNode` now peeks at the configFile via a new `engine.NodeFileHasOverlay` helper (reuses the existing `isEffectivelyEmptyYAML` classifier, so modeline-only / empty / comments-only all return false). When `len(nodes) > 1` and the file carries a body, the call errors with the file, targets, and remediation. Modeline-only multi-node files remain allowed. Regression tests: `TestApplyTemplatesPerNode_MultiNodeWithNonEmptyBodyIsRejected`, `TestApplyTemplatesPerNode_MultiNodeEmptyBodyIsAllowed`, `TestNodeFileHasOverlay`.

README.md
```diff
@@ -140,6 +140,23 @@ Re-template and update generated file in place (this will overwrite it):
...
+> survive even when the template auto-generates conflicting values
+> (e.g. `hostname: talos-XXXXX`).
+>
+> `talm template -f node.yaml` (with or without `-I`) does **not** apply
+> the same overlay: its output is the rendered template plus the modeline
+> and the auto-generated warning, byte-identical to what the template
+> alone would produce. Routing it through the patcher would drop every
+> YAML comment (including the modeline) and re-sort keys, breaking
+> downstream commands that read the file back. Use `apply --dry-run` if
+> you want to preview the exact bytes that will be sent to the node.
```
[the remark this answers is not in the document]
lexfrei: Addressed in 265bff57. Added an explicit v1.12+ caveat that spells out the limitation (`machine.network.interfaces` has no 1:1 mapping to `LinkConfig`/`BondConfig`/`VLANConfig`/`Layer2VIPConfig`) and directs readers to patch the typed resources instead. Also documented the new one-body-one-node guard that landed alongside. Non-network fields (`machine.install.disk`, extra etcd args, etc.) still merge as before — only the legacy network overlay is out of scope.

pkg/commands/apply_test.go
```diff
@@ -247,3 +247,336 @@ func TestWrapWithNodeContext_NoNodesNoClient(t *testing.T) {
...
+
+	// apply path: renders, then overlays.
+	merged, err := engine.MergeFileAsPatch([]byte(rendered), nodeFile)
+	if err != nil {
+		t.Fatalf("MergeFileAsPatch: %v", err)
+	}
+	if !strings.Contains(string(merged), "hostname: node0") {
+		t.Errorf("apply path must overlay hostname: node0; got:\n%s", string(merged))
+	}
+
```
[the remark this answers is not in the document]
lexfrei: Fixed in 294b4cea. You're right — `templateOutput := []byte(rendered)` made both assertions structurally vacuous. Deleted the vacuous block and left a comment explaining that the no-overlay invariant for `talm template` is structural rather than runtime (`pkg/commands/template.go` never calls `engine.MergeFileAsPatch`, and a regression that wired the merge into `generateOutput` would surface in the modeline round-trip tests — the assert-against-the-constant pattern cannot pin it). The apply-path half of the test keeps its meaningful `MergeFileAsPatch` assertion.

pkg/commands/apply.go
```diff
@@ -113,32 +113,41 @@ func apply(args []string) error {
...
 
 		if len(modelineTemplates) > 0 {
-			// Template rendering path: connect to the node first, render templates
-			// online (so lookup() functions resolve real discovery data), then apply.
+			// Template rendering path: render templates online per node and
+			// apply the rendered config plus the node file overlay. See
+			// applyTemplatesPerNode for why the loop is mandatory.
 			opts := buildApplyRenderOptions(modelineTemplates, withSecretsPath)
+			nodes := append([]string(nil), GlobalArgs.Nodes...)
+			fmt.Printf("- talm: file=%s, nodes=%s, endpoints=%s\n", configFile, GlobalArgs.Nodes, GlobalArgs.Endpoints)
```
myasnikovdaniil: **print uses `GlobalArgs.Nodes` while a snapshot exists two lines up**

```go
nodes := append([]string(nil), GlobalArgs.Nodes...)
fmt.Printf("- talm: file=%s, nodes=%s, endpoints=%s\n", configFile, GlobalArgs.Nodes, GlobalArgs.Endpoints)
```

The defensive-copy `nodes` is captured for stable per-iteration reuse, then the print line below it reads `GlobalArgs.Nodes` directly. Functionally identical here (nothing mutates the global between the snapshot and the printf), but it's inconsistent and one inadvertent reorder away from a confusing diagnostic. Suggest using `nodes` for the print:

```go
nodes := append([]string(nil), GlobalArgs.Nodes...)
fmt.Printf("- talm: file=%s, nodes=%s, endpoints=%s\n", configFile, nodes, GlobalArgs.Endpoints)
```

Separately: when the auth path falls back to talosconfig nodes (CLI nodes empty, modeline empty), this line prints `nodes=[]` even though the apply will use the talosconfig context's nodes. This is unchanged from pre-PR behavior, so not a regression — but worth a follow-up to print after `resolveAuthTemplateNodes` resolves.

---

pkg/commands/apply.go
```diff
@@ -188,23 +194,170 @@ func apply(args []string) error {
...
+// captures the GlobalArgs.Nodes value at the moment WithClientMaintenance
+// would have read it.
+func openClientPerNodeMaintenance(fingerprints []string, mkClient maintenanceClientFunc) openClientFunc {
+	return func(node string, action func(ctx context.Context, c *client.Client) error) error {
+		savedNodes := append([]string(nil), GlobalArgs.Nodes...)
+		GlobalArgs.Nodes = []string{node}
+		defer func() { GlobalArgs.Nodes = savedNodes }()
+		return mkClient(fingerprints, action)
+	}
+}
```
myasnikovdaniil, line 322: **global-state mutation as the channel into `WithClientMaintenance`**

```go
func openClientPerNodeMaintenance(fingerprints []string, mkClient maintenanceClientFunc) openClientFunc {
	return func(node string, action func(ctx context.Context, c *client.Client) error) error {
		savedNodes := append([]string(nil), GlobalArgs.Nodes...)
		GlobalArgs.Nodes = []string{node}
		defer func() { GlobalArgs.Nodes = savedNodes }()
		return mkClient(fingerprints, action)
	}
}
```

The narrow-and-restore pattern works (defer fires on panic and on `mkClient` returning an error), and the comment explains why. Two notes:

1. This couples correctness to `WithClientMaintenance` reading `GlobalArgs.Nodes` synchronously inside `mkClient`. It does today (`pkg/talosctl/.../global/client.go` builds the client with `client.WithEndpoints(c.Nodes...)` before calling `action`), but a future Talos bump that defers endpoint resolution (or moves it onto a goroutine) would silently break this. Worth a TODO or, ideally, an upstream `WithClientMaintenanceFor(endpoints []string, ...)` overload that takes endpoints as a parameter.
2. `TestOpenClientPerNodeMaintenance_NarrowsAndRestoresGlobalNodes` only exercises the success path. A one-liner asserting that `GlobalArgs.Nodes` is also restored when `action` returns an error would harden this:

```go
t.Run("restores GlobalArgs.Nodes when action errors", func(t *testing.T) {
	GlobalArgs.Nodes = []string{"orig"}
	defer func() { GlobalArgs.Nodes = nil }()
	openClient := openClientPerNodeMaintenance(nil, fakeMaintenance)
	_ = openClient("10.0.0.1", func(_ context.Context, _ *client.Client) error {
		return fmt.Errorf("boom")
	})
	if !slices.Equal(GlobalArgs.Nodes, []string{"orig"}) {
		t.Errorf("GlobalArgs.Nodes not restored after error path: got %v", GlobalArgs.Nodes)
	}
})
```

Neither point blocks merge.

---

pkg/commands/apply.go
```diff
@@ -188,23 +194,170 @@ func apply(args []string) error {
...
+			)
+		}
+	}
+	for _, node := range nodes {
+		if err := openClient(node, func(ctx context.Context, c *client.Client) error {
+			return renderMergeAndApply(ctx, c, opts, configFile, render, apply)
+		}); err != nil {
+			return fmt.Errorf("node %s: %w", node, err)
+		}
+	}
```
myasnikovdaniil, line 286: **fail-fast on first failing node is a behavioral change**

```go
for _, node := range nodes {
	if err := openClient(node, func(ctx context.Context, c *client.Client) error {
		return renderMergeAndApply(ctx, c, opts, configFile, render, apply)
	}); err != nil {
		return fmt.Errorf("node %s: %w", node, err)
	}
}
```

Previously, batched `ApplyConfiguration` via `apid` would attempt every node and surface aggregated per-node results through `helpers.PrintApplyResults`. The new loop returns on the first failure, so node C is never attempted if node B fails. For a multi-control-plane bootstrap that's probably a footgun: a transient failure on node 2 leaves nodes 3..N unconfigured and the user has to figure out which subset retried. Two reasonable options, in order of effort:

- Document this in the release notes as a deliberate trade-off (the alternative — collect-all-errors — masks the first failure under a multi-error wrapper that's harder to read in the common case).
- Continue past per-node failures, accumulate via `multierror`, print per-node status as you go, and return the aggregate at the end.

Not blocking — flag for the maintainer's call.

---

pkg/engine/engine.go
```diff
@@ -202,6 +204,80 @@ func SerializeConfiguration(configBundle *bundle.Bundle, machineType machine.Typ
...
+		if trimmed[0] == '#' {
+			continue
+		}
+		if string(trimmed) == "---" || string(trimmed) == "..." {
+			continue
+		}
+		return false
+	}
+	return true
+}
```
myasnikovdaniil, line 285: **line-based `isEffectivelyEmptyYAML` over a YAML decoder**

```go
func isEffectivelyEmptyYAML(data []byte) bool {
	for _, line := range bytes.Split(data, []byte("\n")) {
		trimmed := bytes.TrimSpace(line)
		if len(trimmed) == 0 {
			continue
		}
		if trimmed[0] == '#' {
			continue
		}
		if string(trimmed) == "---" || string(trimmed) == "..." {
			continue
		}
		return false
	}
	return true
}
```

Two correctness edge cases worth thinking about:

1. **Inline comments after a key.** A line like `machine: {} # inline` would be classified as non-empty (correct). But a line like `# top-level\nkey: # trailing` — the `key:` line has a non-`#` first character, so it correctly flags non-empty. So this edge is fine.
2. **Indented `---`.** A document separator must be at column 0 per spec. `string(trimmed) == "---"` strips leading whitespace before comparing, so an indented `  ---` (which would actually be a YAML scalar inside a parent mapping) would be misclassified as a separator. In practice nobody writes that, but the comment doc says "comments, document separators, and whitespace" — strictly, this is broader than that.

The simpler alternative: `yaml.NewDecoder` over `data` and call `Decode` until `io.EOF`; if every document yields a `nil` or empty `Node`, return true. That sidesteps both edges and avoids re-implementing a YAML parser. The existing implementation passes the table-driven test, so the bar to switch is low; mention this as a follow-up if you don't want to churn.

---

pkg/engine/engine.go
```diff
@@ -202,6 +204,80 @@ func SerializeConfiguration(configBundle *bundle.Bundle, machineType machine.Typ
...
+	out, err := configpatcher.Apply(configpatcher.WithBytes(rendered), []configpatcher.Patch{patch})
+	if err != nil {
+		return nil, fmt.Errorf("applying patch from %s: %w", patchFile, err)
+	}
+	merged, err := out.Bytes()
+	if err != nil {
+		return nil, fmt.Errorf("encoding merged config from %s: %w", patchFile, err)
+	}
+	return merged, nil
+}
```
myasnikovdaniil, line 246: **`MergeFileAsPatch` re-reads patchFile that was already read by `processModelineAndUpdateGlobals`**

The apply path reads each `configFile` at least three times during one iteration: once in `processModelineAndUpdateGlobals` (modeline parse), once in `engine.NodeFileHasOverlay` (multi-node guard), and once here in `engine.MergeFileAsPatch`. For a normal-sized node file it doesn't matter, but the guard's read could be folded into `MergeFileAsPatch`'s — e.g., have `MergeFileAsPatch` return `(merged, hasOverlay, err)` and let the caller use the `hasOverlay` signal for the multi-node guard before calling apply. Pure cleanup; not blocking.

---

pkg/commands/apply_test.go
```diff
@@ -247,3 +265,415 @@ func TestWrapWithNodeContext_NoNodesNoClient(t *testing.T) {
...
+}
+
+// fakeAuthOpenClient mimics openClientPerNodeAuth for tests: shares one
+// (nil) parent client across iterations and rotates the node via WithNode
+// on a fresh per-iteration context.
+func fakeAuthOpenClient(parentCtx context.Context) openClientFunc {
+	return func(node string, action func(ctx context.Context, c *client.Client) error) error {
+		return action(client.WithNode(parentCtx, node), nil)
+	}
+}
```
myasnikovdaniil, line 301: **`fakeAuthOpenClient` passes a nil `*client.Client`**

```go
func fakeAuthOpenClient(parentCtx context.Context) openClientFunc {
	return func(node string, action func(ctx context.Context, c *client.Client) error) error {
		return action(client.WithNode(parentCtx, node), nil)
	}
}
```

The fake hands `action` a nil `*client.Client`. `applyTemplatesPerNode` doesn't dereference the client itself, but if a future change inside the loop body started calling something on `c` (e.g., `c.GetConfigContext()` for some new fallback), every `fakeAuthOpenClient`-driven test would panic. Consider returning a zero-value sentinel client or changing `action`'s signature in tests to accept a non-nil interface stub. Minor.

---

README.md
```diff
@@ -140,6 +140,42 @@ Re-template and update generated file in place (this will overwrite it):
...
+
 ## Using talosctl commands
```
myasnikovdaniil, line 180: **multi-paragraph blockquote**

Document looks great; one nit: the four `> **bold lead.** ...` paragraphs are easy to skim on GitHub but, in a terminal, the `>` prefix without a blank line between paragraphs renders as one giant quote. Adding `>` blank lines (already present) is fine; mainly flagging that the structure is dense and could land as a nested bullet list under one `> **Per-node patches inside node files.**` heading.

Not actionable; just an observation.

---

myasnikovdaniil (commented):
Now I have enough context. Drafting the review.

---

# PR 128 Review Draft

## Summary

This PR fixes a real, shipping regression introduced by #119: the template-rendering branch of `talm apply` batched all nodes into a single gRPC context, which `engine.Render` rejected via `FailIfMultiNodes` before any rendering happened. The fix iterates per node — single-node ctx in auth mode (`client.WithNode`) and a fresh single-endpoint client in maintenance mode (because `WithClientMaintenance` ignores per-call node metadata and would round-robin `ApplyConfiguration` across all endpoints otherwise). The companion fixes (overlay-merge of the node file body via `engine.MergeFileAsPatch`, error-message subcommand thread-through, error-handling symmetry, and the dead `Options.Insecure` removal) are tightly scoped and all sit on the same callback. Tests are comprehensive and the README clearly documents the apply/template divergence and the Talos v1.12+ caveat. I found no blockers; comments below are mostly nits.

### Nit — test names

Tests like `TestApplyTemplatesPerNode_LoopsOncePerNodeWithSingleNodeContext` and `TestOpenClientPerNodeMaintenance_NarrowsAndRestoresGlobalNodes` are descriptive but very long. `t.Run` subtests with shorter parent names would read better in `go test -v` output. Style call.

---

## Verification I performed

- Traced `client.WithNode` (`pkg/machinery/client/context.go:30`) — confirmed it deletes the `nodes` md key and sets `node`, so `WithClientSkipVerify`'s pre-injected multi-node ctx is correctly narrowed by `openClientPerNodeAuth`. The `TestApplyTemplatesPerNode_AuthModeUsesSingleNodeMetadataKey` test validates this.
- Traced `helpers.FailIfMultiNodes` — confirmed it only inspects the `nodes` md key, not `node`, so single-target metadata passes the guard trivially.
- Traced `configpatcher.LoadPatch` and `JSON6902` — confirmed an effectively-empty patch is decoded as an empty `jsonpatch.Patch`, and `JSON6902` then rejects multi-doc input with `JSON6902 patches are not supported for multi-document machine configuration`. The empty-content short-circuit in `MergeFileAsPatch` is genuinely necessary for the v1.12+ multi-doc default output.
- Traced `Args.WithClientMaintenance` — confirmed it reads `c.Nodes` (= `GlobalArgs.Nodes`) synchronously inside the same goroutine before calling `action`, so the narrow-and-restore pattern is correct under today's Talos client semantics.

## Recommendation

**Comment** — approve once the maintainer decides on the fail-fast-vs-multierror call (or chooses to document it and move on). Everything else is cleanup-grade.

pkg/commands/apply.go
lexfrei: Switched the printf to use the snapshotted `nodes` in 6d1339b. Agreed it was one inadvertent reorder away from a confusing diagnostic.

Re: the post-`resolveAuthTemplateNodes` print — kept that out of scope as you suggested; the current line still shows what the user asked for (CLI/modeline) before the talosconfig fallback resolves, which is the more useful signal when the print fires before the fallback. Happy to fold a "resolved nodes" follow-up print into a separate PR if you'd rather see both.

pkg/commands/apply.go
lexfrei: Going with fail-fast as the deliberate behavior, and I'll spell that out in the PR description rather than rely on the reader to infer it from the loop. The reasoning: for a multi-control-plane bootstrap, continuing past a node-2 failure to apply node-3 risks driving the cluster's quorum into a worse state than stopping at the first failure does — the operator has to retry node 2 either way, but a partial-but-uniform application is easier to reason about than a partial-and-skipping one. The aggregate-and-continue alternative is reasonable for a `talm apply --best-effort`-style flag, which is a clean follow-up: the loop here would gain a `multierror` accumulator behind a flag default-false, leaving today's fail-fast as the default. Filed mentally; will open if needed.

pkg/commands/apply.go
lexfrei: Both notes addressed.

Note 1 — added a paragraph to `openClientPerNodeMaintenance`'s package doc in 6d1339b spelling out the synchronous-read assumption: the narrowing only works because `mkClient` reads `GlobalArgs.Nodes` before `action` returns. A future Talos refactor that defers endpoint resolution onto a goroutine would silently break this. Took the doc-it-now path rather than chase an upstream `WithClientMaintenanceFor(endpoints, ...)` overload in this PR.

Note 2 — added `TestOpenClientPerNodeMaintenance_RestoresGlobalNodesOnError` in 6403d1e, mirroring the t.Run sketch you wrote out. Pinning the error path next to the success path means a future swap of the `defer` for an explicit "restore after success" block would regress audibly.

pkg/commands/apply_test.go
lexfrei: Documented the contract in 6403d1e rather than swap to a zero-value `&client.Client{}`. Reasoning: the nil here is deliberate — if a future change inside `applyTemplatesPerNode` starts dereferencing `c`, every fake-driven test panics in CI immediately. A zero-value would change "panic on call" into a different, possibly silent failure mode (the talos client struct has internal state that zero-init does not satisfy). The comment now spells this out so the design intent is discoverable.

pkg/engine/engine.go
lexfrei: Pure cleanup as you say. Declining to fold this into the PR: changing `MergeFileAsPatch`'s signature from `(merged, error)` to `(merged, hasOverlay, error)` widens the API for every caller, and the redundant read costs ~µs on a node-file-sized input. Happy to revisit if a profile shows it pays off, but absent that, keeping the small, single-purpose function. Filed mentally as a follow-up signal rather than work.

pkg/engine/engine.go
lexfrei: Fixed the indented-separator edge in c5555c2 — separator comparison now uses `bytes.TrimRight` (trailing whitespace only) rather than `bytes.TrimSpace`, so an indented `  ---` no longer matches. Added an "indented separator counts as overlay" case to `TestNodeFileHasOverlay`.

Stuck with the line-based scan over `yaml.NewDecoder` for now: the function is small, the existing table tests cover the contract, and a real YAML decoder pulls in a heavier dependency for a path that runs once per apply. The edge you flagged was the genuinely incorrect case; the rest is in the "nobody writes that" zone.

README.md
lexfrei: Noted — agreed the four bold-lead paragraphs are easier to skim on GitHub than in a terminal. Will refactor as a nested bullet list under one `> **Per-node patches inside node files.**` heading next time I touch this section. Leaving the current shape this PR.

myasnikovdaniil (approved):
All review comments addressed across commits 6d1339b, 6403d1e5, c5555c2a; fail-fast behavior decision documented. LGTM.

PR #129, opened 2026-04-17
fix(windows): CI matrix, zip archives, and NTFS ACL for sensitive files

### LINKED ISSUE #11: Windows support is broken
Same issue with #10, but it is more of a general issue with Windows

lexfrei:
## Summary

Close the Windows support loop. The original "template path with backslash" symptom was already fixed upstream of this branch (engine-side `NormalizeTemplatePath` in v0.22.0), but nothing proved it stayed fixed — CI only ran on Ubuntu, so the Windows-tagged tests never executed, and the release archive was a `.tar.gz` that stock Windows cannot open. This PR makes Windows a first-class target and closes the security gap that becomes visible once Windows users can actually run talm.

Closes #11

### 1. Windows CI runner

`.github/workflows/pr.yml` runs `go test ./...` on both `ubuntu-latest` and `windows-latest` with `fail-fast: false`. The existing `engine_windows_test.go` (build-tagged `//go:build windows`) now actually executes. Any future regression in Windows-specific path handling fails CI.

### 2. Windows zip archives

`.goreleaser.yaml` uses `format_overrides` so Windows binaries ship as `.zip`. Linux and macOS remain `.tar.gz`.

### 3. Regression tests for backslash paths

Users running talm from PowerShell pass paths with `\` separators, which must be normalized before reaching the helm engine's forward-slash-only map keys. Both code paths that touch template arguments now have Windows-tagged tests:

- `apply_windows_test.go::TestResolveTemplatePaths_BackslashInput` covers `talm apply` via `resolveTemplatePaths`.
- `template_windows_test.go::TestResolveEngineTemplatePaths_BackslashInput` covers `talm template` via the newly-extracted `resolveEngineTemplatePaths`.

The two resolvers are intentionally different (apply resolves against `rootDir`, template against CWD with a `templates/<basename>` fallback), so they need separate coverage.

### 4. New `pkg/secureperm` package for sensitive files

A drop-in replacement for `os.WriteFile(path, data, 0o600)` and `os.Chmod(path, 0o600)`. On Unix it delegates to those calls with an explicit follow-up `Chmod` (required because `os.WriteFile` preserves prior mode bits on overwrite). On Windows it installs a protected NTFS DACL with a single Allow ACE for the current user SID.

WriteFile is also **atomic** on both platforms via write-to-tmp + rename — a failure mid-write leaves the original file intact. Secrets files are not reconstructible (corrupting `secrets.yaml` forces a cluster PKI reissue), so non-atomic writes are a real hazard.

Call sites migrated:

- `pkg/age/age.go` — `talm.key` creation, decrypted `secrets.yaml`, generic decrypted YAML.
- `pkg/commands/init.go` — talosconfig, secrets.yaml bundle (via the new `writeSecureToDestination` helper).
- `pkg/commands/rotate_ca_handler.go` — rotated `secrets.yaml`.
- `pkg/commands/talosconfig.go` — regenerated talosconfig.
- `pkg/commands/kubeconfig_handler.go` — kubeconfig tighten-after-fetch.
- `pkg/commands/template.go` — rendered machine config via `talm template --inplace` (embeds certs + PKI keys + cluster join tokens).

Tests cover the atomic-write contract on both OSes (`TestWriteFile_PreservesOriginalOnFailure_{Unix,Windows}`), the DACL shape on Windows (`TestWriteFile_*_DACL_Windows` parse the SDDL string and assert `D:P` + exactly one Allow ACE for the current user SID), overwrite-downgrade behavior on Unix (`TestWriteFile_OverwriteDowngrades_Unix`), and the new-file happy path on both.

### 5. Collateral cleanups

Three small defects surfaced while working in this area:

- `pkg/commands/kubeconfig_handler.go` had a vacuous `if err == nil { ... }` wrapper around the entire 60-line post-write block, guarding a condition that was already checked with early-return one line above. The wrapper also shadowed `err` with inner `filepath.Abs` assignments, making the outer variable meaningless. Removed.
- `pkg/commands/init.go::writeToDestination` printed `Created <path>` unconditionally after the write, so a permission failure produced a confusing "Created foo.key" + error pair. Both `writeToDestination` and the new `writeSecureToDestination` now gate the message on `err == nil`. Pinned by `TestWriteToDestination_SilentOnFailure` and `TestWriteToDestination_AnnouncesOnSuccess`.
- `writeSecretsBundleToFile` called `validateFileExists` twice — the outer call was redundant with the same check inside `writeSecureToDestination`. Removed.

### 6. README

New "Windows" installation subsection pointing at the release zip and documenting the `-t` / `--template` separator equivalence.

## Verification

- `go test ./...` on macOS — pass
- `GOOS=windows GOARCH=amd64 go build ./...` — clean
- `GOOS=windows GOARCH=amd64 go vet ./...` — clean
- `GOOS=windows GOARCH=amd64 go test -c -o /dev/null ./...` — all test packages compile for Windows
- `golangci-lint run ./...` — 0 issues
- CI matrix (`ubuntu-latest` + `windows-latest`) is exercised by this PR itself — watch for `--- PASS` on both runners

## Summary by CodeRabbit

* **New Features**
  * Windows platform support with installation docs and Windows-compatible template path handling.
  * Secure, cross-platform writing for sensitive outputs to ensure owner-only access.

* **Bug Fixes**
  * Consistent permission/ACL tightening for secret files across platforms.

* **Tests**
  * Expanded Unix and Windows tests covering permission, overwrite, and path-resolution behaviors.

* **Chores**
  * CI matrix expanded to include Windows; release/archive formats and dependency declarations updated.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -7,7 +7,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@v6
--- a/pkg/commands/init.go
+++ b/pkg/commands/init.go
@@ -345,7 +345,7 @@ var initCmd = &cobra.Command{
 				return fmt.Errorf("failed to marshal config: %+v", err)
 			}
 
-			if err = writeToDestination(data, talosconfigFile, 0o600); err != nil {
+			if err = writeSecureToDestination(data, talosconfigFile); err != nil {
 				return err
 			}
 		}
@@ -459,7 +459,7 @@ func writeSecretsBundleToFile(bundle *secrets.Bundle) error {
 		return err
 	}
 
-	return writeToDestination(bundleBytes, secretsFile, 0o600)
+	return writeSecureToDestination(bundleBytes, secretsFile)
 }
 
 // readChartYamlPreset reads Chart.yaml and determines the preset name from dependencies
@@ -874,15 +874,29 @@ func writeToDestination(data []byte, destination string, permissions os.FileMode
 		return fmt.Errorf("failed to create output dir: %w", err)
 	}
 
-	// Route owner-only writes through secureperm so Windows files get an
-	// NTFS DACL applied — os.WriteFile's mode argument is effectively
-	// ignored on Windows.
-	var err error
-	if permissions == 0o600 {
-		err = secureperm.WriteFile(destination, data)
-	} else {
-		err = os.WriteFile(destination, data, permissions)
+	err := os.WriteFile(destination, data, permissions)
+
+	fmt.Fprintf(os.Stderr, "Created %s\n", destination)
+
+	return err
+}
+
+// writeSecureToDestination writes a secret (talosconfig, secrets.yaml,
+// talm.key) with owner-only permissions. On Windows the NTFS DACL is
+// installed via secureperm so os.WriteFile's ignored mode bits aren't
+// the only defense.
+func writeSecureToDestination(data []byte, destination string) error {
+	if err := validateFileExists(destination); err != nil {
+		return err
 	}
+
+	parentDir := filepath.Dir(destination)
+
+	if err := os.MkdirAll(parentDir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create output dir: %w", err)
+	}
+
+	err := secureperm.WriteFile(destination, data)
 
 	fmt.Fprintf(os.Stderr, "Created %s\n", destination)
 
--- a/pkg/commands/init.go
+++ b/pkg/commands/init.go
@@ -456,10 +456,8 @@ func writeSecretsBundleToFile(bundle *secrets.Bundle) error {
 	}
 
 	secretsFile := filepath.Join(Config.RootDir, "secrets.yaml")
-	if err = validateFileExists(secretsFile); err != nil {
-		return err
-	}
-
+	// validateFileExists is invoked inside writeSecureToDestination;
+	// no need to duplicate the --force / existing-file gate here.
 	return writeSecureToDestination(bundleBytes, secretsFile)
 }
 
--- a/pkg/commands/init.go
+++ b/pkg/commands/init.go
@@ -895,7 +895,10 @@ func writeSecureToDestination(data []byte, destination string) error {
 
 	parentDir := filepath.Dir(destination)
 
-	if err := os.MkdirAll(parentDir, os.ModePerm); err != nil {
+	// Use 0o700 so any newly-created parent dir for secrets is owner-only
+	// even under a permissive umask. MkdirAll is a no-op when the dir
+	// already exists, so this does not override pre-existing dir perms.
+	if err := os.MkdirAll(parentDir, 0o700); err != nil {
 		return fmt.Errorf("failed to create output dir: %w", err)
 	}
 
--- a/pkg/commands/init.go
+++ b/pkg/commands/init.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/cozystack/talm/pkg/age"
 	"github.com/cozystack/talm/pkg/generated"
+	"github.com/cozystack/talm/pkg/secureperm"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 
@@ -873,7 +874,15 @@ func writeToDestination(data []byte, destination string, permissions os.FileMode
 		return fmt.Errorf("failed to create output dir: %w", err)
 	}
 
-	err := os.WriteFile(destination, data, permissions)
+	// Route owner-only writes through secureperm so Windows files get an
+	// NTFS DACL applied — os.WriteFile's mode argument is effectively
+	// ignored on Windows.
+	var err error
+	if permissions == 0o600 {
+		err = secureperm.WriteFile(destination, data)
+	} else {
+		err = os.WriteFile(destination, data, permissions)
+	}
 
 	fmt.Fprintf(os.Stderr, "Created %s\n", destination)
 
--- a/pkg/commands/init.go
+++ b/pkg/commands/init.go
@@ -17,6 +17,7 @@ package commands
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"slices"
@@ -862,6 +863,10 @@ func handleTalosconfigEncryption(requireKeyForDecrypt bool) (bool, error) {
 	return keyWasCreated, nil
 }
 
+// createdSink is where "Created <path>" messages go after a successful
+// write. Swappable in tests to assert no message is emitted on failure.
+var createdSink io.Writer = os.Stderr
+
 func writeToDestination(data []byte, destination string, permissions os.FileMode) error {
 	if err := validateFileExists(destination); err != nil {
 		return err
@@ -875,9 +880,9 @@ func writeToDestination(data []byte, destination string, permissions os.FileMode
 	}
 
 	err := os.WriteFile(destination, data, permissions)
-
-	fmt.Fprintf(os.Stderr, "Created %s\n", destination)
-
+	if err == nil {
+		_, _ = fmt.Fprintf(createdSink, "Created %s\n", destination)
+	}
 	return err
 }
 
@@ -897,8 +902,8 @@ func writeSecureToDestination(data []byte, destination string) error {
 	}
 
 	err := secureperm.WriteFile(destination, data)
-
-	fmt.Fprintf(os.Stderr, "Created %s\n", destination)
-
+	if err == nil {
+		_, _ = fmt.Fprintf(createdSink, "Created %s\n", destination)
+	}
 	return err
 }
--- a/pkg/commands/kubeconfig_handler.go
+++ b/pkg/commands/kubeconfig_handler.go
@@ -101,62 +101,60 @@ Otherwise, kubeconfig will be written to PWD.`
 			return fmt.Errorf("failed to resolve kubeconfig path: %w", err)
 		}
 
-		if err == nil {
-			// Set secure permissions (600) on kubeconfig file. On Windows
-			// this lays down an NTFS DACL; os.Chmod would have been a no-op.
-			if err := secureperm.LockDown(absPath); err != nil {
-				// Don't fail the command if the tighten fails, but log warning
-				fmt.Fprintf(os.Stderr, "Warning: failed to set permissions on kubeconfig: %v\n", err)
-			}
+		// Set secure permissions (600) on kubeconfig file. On Windows
+		// this lays down an NTFS DACL; os.Chmod would have been a no-op.
+		if err := secureperm.LockDown(absPath); err != nil {
+			// Don't fail the command if the tighten fails, but log warning
+			fmt.Fprintf(os.Stderr, "Warning: failed to set permissions on kubeconfig: %v\n", err)
+		}
 
-			rootAbs, err := filepath.Abs(Config.RootDir)
-			if err == nil {
-				relPath, err := filepath.Rel(rootAbs, absPath)
-				if err == nil && !strings.HasPrefix(relPath, "..") {
-					// Path is within project root, add to .gitignore
-					fileName := filepath.Base(kubeconfigPath)
-					if fileName == "kubeconfig" {
-						if err := addToGitignore("kubeconfig"); err != nil {
-							// Don't fail the command if gitignore update fails
-							fmt.Fprintf(os.Stderr, "Warning: failed to update .gitignore: %v\n", err)
-						}
+		rootAbs, err := filepath.Abs(Config.RootDir)
+		if err == nil {
+			relPath, err := filepath.Rel(rootAbs, absPath)
+			if err == nil && !strings.HasPrefix(relPath, "..") {
+				// Path is within project root, add to .gitignore
+				fileName := filepath.Base(kubeconfigPath)
+				if fileName == "kubeconfig" {
+					if err := addToGitignore("kubeconfig"); err != nil {
+						// Don't fail the command if gitignore update fails
+						fmt.Fprintf(os.Stderr, "Warning: failed to update .gitignore: %v\n", err)
 					}
 				}
 			}
+		}
 
-			// Update kubeconfig server endpoint if endpoint is available
-			if len(GlobalArgs.Endpoints) > 0 {
-				endpoint := GlobalArgs.Endpoints[0]
-				if err := updateKubeconfigServer(absPath, endpoint); err != nil {
-					// Don't fail the command if update fails, but log warning
-					fmt.Fprintf(os.Stderr, "Warning: failed to update kubeconfig server endpoint: %v\n", err)
-				}
+		// Update kubeconfig server endpoint if endpoint is available
+		if len(GlobalArgs.Endpoints) > 0 {
+			endpoint := GlobalArgs.Endpoints[0]
+			if err := updateKubeconfigServer(absPath, endpoint); err != nil {
+				// Don't fail the command if update fails, but log warning
+				fmt.Fprintf(os.Stderr, "Warning: failed to update kubeconfig server endpoint: %v\n", err)
 			}
+		}
 
-			// Automatically update kubeconfig.encrypted if it exists and talm.key exists
-			// Skip this if --login flag is set
-			if !loginFlagValue {
-				// Get relative path from project root for encryption
-				rootAbs, err := filepath.Abs(Config.RootDir)
-				if err == nil {
-					relKubeconfigPath, err := filepath.Rel(rootAbs, absPath)
-					if err == nil && !strings.HasPrefix(relKubeconfigPath, "..") {
-						// Path is within project root
-						encryptedKubeconfigPath := relKubeconfigPath + ".encrypted"
-						encryptedKubeconfigFile := filepath.Join(Config.RootDir, encryptedKubeconfigPath)
-						keyFile := filepath.Join(Config.RootDir, "talm.key")
+		// Automatically update kubeconfig.encrypted if it exists and talm.key exists
+		// Skip this if --login flag is set
+		if !loginFlagValue {
+			// Get relative path from project root for encryption
+			rootAbs, err := filepath.Abs(Config.RootDir)
+			if err == nil {
+				relKubeconfigPath, err := filepath.Rel(rootAbs, absPath)
+				if err == nil && !strings.HasPrefix(relKubeconfigPath, "..") {
+					// Path is within project root
+					encryptedKubeconfigPath := relKubeconfigPath + ".encrypted"
+					encryptedKubeconfigFile := filepath.Join(Config.RootDir, encryptedKubeconfigPath)
+					keyFile := filepath.Join(Config.RootDir, "talm.key")
 
-						encryptedExists := fileExists(encryptedKubeconfigFile)
-						keyExists := fileExists(keyFile)
+					encryptedExists := fileExists(encryptedKubeconfigFile)
+					keyExists := fileExists(keyFile)
 
-						if encryptedExists && keyExists {
-							// Both files exist, encrypt kubeconfig
-							if err := age.EncryptYAMLFile(Config.RootDir, relKubeconfigPath, encryptedKubeconfigPath); err != nil {
-								// Don't fail the command if encryption fails, but log warning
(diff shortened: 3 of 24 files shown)


pkg/commands/init.go
```diff
@@ -874,8 +878,30 @@ func writeToDestination(data []byte, destination string, permissions os.FileMode
...
+// installed via secureperm so os.WriteFile's ignored mode bits aren't
+// the only defense.
+func writeSecureToDestination(data []byte, destination string) error {
+	if err := validateFileExists(destination); err != nil {
+		return err
+	}
+
+	parentDir := filepath.Dir(destination)
+
+	if err := os.MkdirAll(parentDir, os.ModePerm); err != nil {
```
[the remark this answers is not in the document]
lexfrei: Switched to `0o700` in cddfef47. `MkdirAll` is a no-op when the directory already exists, so this only affects freshly created parent dirs for secrets — the files themselves were already 0o600 via `secureperm`, but the parent could land at 0o755 under a permissive umask, which didn't match the helper's goal.

pkg/commands/template_unix_test.go
```diff
@@ -0,0 +1,70 @@
...
+// file at 0o644; on update the user expects the secret bits to be
+// tightened, not preserved.
+func TestWriteInplaceRendered_OverwriteDowngrades_Unix(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, "node0.yaml")
+
+	if err := os.WriteFile(path, []byte("old"), 0o644); err != nil {
+		t.Fatalf("seed: %v", err)
+	}
+	if err := writeInplaceRendered(path, "new"); err != nil {
```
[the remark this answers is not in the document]
lexfrei: Added an explicit `os.Chmod(path, 0o644)` after the seed in a1a71eda so the lax-mode precondition is deterministic regardless of the developer's umask.

pkg/commands/template.go
```diff
@@ -415,3 +366,73 @@ func init() {
...
+			possiblePath := filepath.Join("templates", templateName)
+			fullPath := filepath.Join(absRootDir, possiblePath)
+			if _, statErr := os.Stat(fullPath); statErr == nil {
+				relPath = possiblePath
+			} else {
+				resolved[i] = engine.NormalizeTemplatePath(templatePath)
+				continue
+			}
+		}
+		resolved[i] = engine.NormalizeTemplatePath(relPath)
```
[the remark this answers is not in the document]
lexfrei: Good catch — fixed in 1df71183. Extracted an `isOutsideRoot` helper that matches the exact `..` path element or `..` + `filepath.Separator`, and routed all three duplicate check sites through it (`apply.go` `resolveTemplatePaths`, `template.go` `generateOutput` modeline branch, `template.go` `resolveEngineTemplatePaths`). Added a regression test that seeds a real `rootDir/..templates/controlplane.yaml` alongside a `rootDir/templates/controlplane.yaml` decoy and asserts the real file is returned; on the old check the test surfaces the decoy via the basename fallback.

pkg/secureperm/secureperm_unix_test.go
```diff
@@ -0,0 +1,126 @@
...
+// atomic tmp+rename strategy achieves this because the renamed tmp
+// file was created with 0o600.
+func TestWriteFile_OverwriteDowngrades_Unix(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, "lax.txt")
+
+	if err := os.WriteFile(path, []byte("old"), 0o644); err != nil {
+		t.Fatalf("seed: %v", err)
+	}
+	if err := secureperm.WriteFile(path, []byte("new")); err != nil {
```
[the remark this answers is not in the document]
lexfrei: Chmod-after-seed added to both `TestLockDown_Mode0600_Unix` and `TestWriteFile_OverwriteDowngrades_Unix` in f4d69323 for the same umask-independence reason. `TestWriteFile_PreservesOriginalOnFailure_Unix` seeds at 0o600 on purpose (it asserts the existing content survives), so it doesn't need the chmod.

pkg/secureperm/secureperm_unix.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	if err := f.Close(); err != nil {
+		return fmt.Errorf("close tmp: %w", err)
+	}
+	if err := os.Rename(tmpPath, path); err != nil {
+		return fmt.Errorf("rename tmp -> %s: %w", path, err)
+	}
+	committed = true
+	return nil
+}
```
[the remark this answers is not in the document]
lexfrei: Fair — added in 86909935. `f.Sync()` runs between `Write` and `Close` and returns any error because a failed fsync means the payload isn't on disk yet. After the rename, the parent directory is opened and best-effort synced; errors are ignored there because a handful of filesystems don't support dir fsync and the tmp fsync already covers the payload itself.

pkg/secureperm/secureperm_windows.go
```diff
@@ -0,0 +1,196 @@
...
+			_ = os.Remove(tmpPath)
+		}
+	}()
+
+	if _, err := f.Write(data); err != nil {
+		return fmt.Errorf("write tmp %s: %w", tmpPath, err)
+	}
+	if err := f.Close(); err != nil {
+		return fmt.Errorf("close tmp %s: %w", tmpPath, err)
+	}
```
myasnikovdaniil, line 179: **missing fsync before rename on Windows**

The Unix variant explicitly sync's the tmp before close + sync's the parent dir after rename, with this rationale:

> fsync the tmp file before rename so its contents are on stable storage; otherwise a crash/power-loss between rename and the delayed disk flush can surface the renamed inode pointing at zero-length or stale data on reboot — the canonical failure mode the atomic-rename pattern is meant to avoid. Secrets files are not reconstructible, so the full fsync is warranted.

The Windows variant skips this entirely. `os.Rename` on Windows uses `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` *without* `MOVEFILE_WRITE_THROUGH`, so a power loss between rename and the OS cache flush leaves you exactly in the failure mode the Unix doc warns about. `*os.File.Sync()` on Windows calls `FlushFileBuffers` and is the right tool here.

```go
	if _, err := f.Write(data); err != nil {
		return fmt.Errorf("write tmp %s: %w", tmpPath, err)
	}
	if err := f.Sync(); err != nil {
		return fmt.Errorf("sync tmp %s: %w", tmpPath, err)
	}
	if err := f.Close(); err != nil {
		return fmt.Errorf("close tmp %s: %w", tmpPath, err)
	}
	if err := os.Rename(tmpPath, path); err != nil {
		return fmt.Errorf("rename tmp %s -> %s: %w", tmpPath, path, err)
	}
	committed = true
```

If you want full parity with Unix you'd also want a directory flush after rename, though Windows has no clean public API for that and most filesystems don't need it — the file flush is the more important of the two.

---

pkg/commands/kubeconfig_handler.go
```diff
@@ -100,61 +101,60 @@ Otherwise, kubeconfig will be written to PWD.`
...
-					fileName := filepath.Base(kubeconfigPath)
-					if fileName == "kubeconfig" {
-						if err := addToGitignore("kubeconfig"); err != nil {
-							// Don't fail the command if gitignore update fails
-							fmt.Fprintf(os.Stderr, "Warning: failed to update .gitignore: %v\n", err)
-						}
+		rootAbs, err := filepath.Abs(Config.RootDir)
+		if err == nil {
+			relPath, err := filepath.Rel(rootAbs, absPath)
+			if err == nil && !strings.HasPrefix(relPath, "..") {
```
myasnikovdaniil: **same `HasPrefix("..")` misclassification this PR fixes elsewhere**

This PR adds `isOutsideRoot` to fix exactly this bug in `apply.go` and `template.go`, but `kubeconfig_handler.go` (in the same package) still has two raw `strings.HasPrefix(relPath, "..")` checks. A user with a sibling directory like `..mykube` would have their kubeconfig misclassified as outside-root and the gitignore-add / encrypted-update branches would be skipped silently.

The impact is smaller than the template-resolution case (it just suppresses two side-effects rather than producing a "template not found"), but it's the same logical bug and the helper to fix it is already in the same package:

```go
		relPath, err := filepath.Rel(rootAbs, absPath)
		if err == nil && !isOutsideRoot(relPath) {
			// Path is within project root, add to .gitignore
```

and

```go
				relKubeconfigPath, err := filepath.Rel(rootAbs, absPath)
				if err == nil && !isOutsideRoot(relKubeconfigPath) {
					// Path is within project root
```

Worth either folding into this PR (it's a one-line change in two places) or filing a follow-up so it doesn't drift.

---

pkg/commands/kubeconfig_handler.go
```diff
@@ -100,61 +101,60 @@ Otherwise, kubeconfig will be written to PWD.`
...
-								fmt.Fprintf(os.Stderr, "Updated %s\n", encryptedKubeconfigPath)
-							}
+		// Automatically update kubeconfig.encrypted if it exists and talm.key exists
+		// Skip this if --login flag is set
+		if !loginFlagValue {
+			// Get relative path from project root for encryption
+			rootAbs, err := filepath.Abs(Config.RootDir)
+			if err == nil {
+				relKubeconfigPath, err := filepath.Rel(rootAbs, absPath)
+				if err == nil && !strings.HasPrefix(relKubeconfigPath, "..") {
```
myasnikovdaniil: **same `HasPrefix("..")` misclassification this PR fixes elsewhere**

This PR adds `isOutsideRoot` to fix exactly this bug in `apply.go` and `template.go`, but `kubeconfig_handler.go` (in the same package) still has two raw `strings.HasPrefix(relPath, "..")` checks. A user with a sibling directory like `..mykube` would have their kubeconfig misclassified as outside-root and the gitignore-add / encrypted-update branches would be skipped silently.

The impact is smaller than the template-resolution case (it just suppresses two side-effects rather than producing a "template not found"), but it's the same logical bug and the helper to fix it is already in the same package:

```go
		relPath, err := filepath.Rel(rootAbs, absPath)
		if err == nil && !isOutsideRoot(relPath) {
			// Path is within project root, add to .gitignore
```

and

```go
				relKubeconfigPath, err := filepath.Rel(rootAbs, absPath)
				if err == nil && !isOutsideRoot(relKubeconfigPath) {
					// Path is within project root
```

Worth either folding into this PR (it's a one-line change in two places) or filing a follow-up so it doesn't drift.

---

pkg/secureperm/secureperm_windows.go
```diff
@@ -0,0 +1,196 @@
...
+// MoveFileEx with MOVEFILE_REPLACE_EXISTING under the hood (os.Rename
+// on Windows), which preserves the tmp's owner-only DACL on the final
+// file. At no point do the new bytes exist on disk under a permissive
+// DACL, and the old bytes remain readable by the owner until the final
+// rename succeeds.
+func WriteFile(path string, data []byte) error {
+	dir := filepath.Dir(path)
+	if dir == "" {
+		dir = "."
+	}
```
myasnikovdaniil: **dead guard**

```go
	dir := filepath.Dir(path)
	if dir == "" {
		dir = "."
	}
```

`filepath.Dir` is documented to never return the empty string — it returns `"."` for a bare filename and `"."` for an empty input. The guard never fires. Either drop it or leave a comment explaining what it's defending against; right now it reads like there's a code path that produces `""` and there isn't.

---

pkg/secureperm/secureperm_unix.go
```diff
@@ -0,0 +1,90 @@
...
+	committed = true
+	// Best-effort fsync of the parent dir so the rename entry itself is
+	// durable. Ignored errors: dir fsync is unsupported on a few
+	// filesystems; the tmp fsync above already protects the payload.
+	if d, openErr := os.Open(dir); openErr == nil {
+		_ = d.Sync()
+		_ = d.Close()
+	}
+	return nil
+}
```
myasnikovdaniil, line 94: **ownership change when run as a different user from the prior owner**

Worth calling out in the package doc. With the old `os.WriteFile(path, data, 0o600)` against an existing file, the inode is opened with `O_TRUNC` so owner/group are preserved. With the new tmp + rename strategy, the result is owned by whoever runs the process. If a user ever runs `talm` as root (e.g. via `sudo talm rotate-ca` on a CI box that had previously been operated by an unprivileged user), the secrets files will silently change ownership to root and the original user can't read them anymore.

Probably fine for the typical single-user workstation flow, but the contract change deserves one line in the package doc next to the atomicity claim. No code change required if you're comfortable with the trade-off.

---

pkg/commands/template_test.go
```diff
@@ -0,0 +1,69 @@
...
+	rootDir, err := filepath.EvalSymlinks(t.TempDir())
+	if err != nil {
+		t.Fatalf("eval symlinks: %v", err)
+	}
+	// Seed both ..templates/controlplane.yaml (the real target) and
+	// templates/controlplane.yaml (a decoy the buggy fallback would
+	// have picked instead).
+	if err := os.MkdirAll(filepath.Join(rootDir, "..templates"), 0o755); err != nil {
+		t.Fatalf("mkdir ..templates: %v", err)
+	}
```
myasnikovdaniil, line 44: **`os.Chdir` in tests**

Both new tests do `os.Chdir(rootDir)` with `t.Cleanup` to restore. `os.Chdir` is process-global, so anyone adding `t.Parallel()` to these or to a sibling test would silently corrupt state. The existing `t.Cleanup` keeps them honest in serial mode, but a `// not safe with t.Parallel — uses os.Chdir` comment would prevent a future regression.

---

pkg/commands/template_windows_test.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := os.MkdirAll(filepath.Join(absRoot, "templates"), 0o755); err != nil {
+		t.Fatalf("mkdir: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(absRoot, "templates", "controlplane.yaml"), []byte{}, 0o600); err != nil {
+		t.Fatalf("seed: %v", err)
+	}
+
+	// Chdir so relative paths resolve against rootDir exactly as they
+	// would when a user cd's into their project directory and runs
+	// `talm template -t templates\controlplane.yaml`.
```
myasnikovdaniil, line 62: **`os.Chdir` in tests**

Both new tests do `os.Chdir(rootDir)` with `t.Cleanup` to restore. `os.Chdir` is process-global, so anyone adding `t.Parallel()` to these or to a sibling test would silently corrupt state. The existing `t.Cleanup` keeps them honest in serial mode, but a `// not safe with t.Parallel — uses os.Chdir` comment would prevent a future regression.

---

.github/workflows/pr.yml
```diff
@@ -7,7 +7,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
```
myasnikovdaniil, line 14: **`fail-fast: false` is correct, just confirming intent**

```yaml
    strategy:
      fail-fast: false
      matrix:
        os: [ubuntu-latest, windows-latest]
    runs-on: ${{ matrix.os }}
```

Reads as "let both runners finish even if one fails." That's the right call for a platform-parity matrix — you want to see Windows-specific failures even if Linux passed (or vice versa). Just calling it out so the PR description's "watch for `--- PASS` on both runners" claim is verified by behavior, not just name.

---

myasnikovdaniil (commented):
I have enough to draft the review. Here it is.

---

# Draft review for PR #129

**Title:** `fix(windows): CI matrix, zip archives, and NTFS ACL for sensitive files`
**Author:** lexfrei
**Base:** main · **Head:** fix/windows-support · **Files:** 24

## Summary

This is a substantial and well-executed Windows-support PR. The CI matrix + zip archives convert Windows from a theoretical to a real release target; the new `pkg/secureperm` package replaces scattered `os.WriteFile(..., 0o600)` / `os.Chmod(..., 0o600)` calls — which were silently no-ops for permissions on NTFS — with an atomic write-to-tmp + rename that lays down a protected owner-only DACL on Windows and an explicit `Chmod` on Unix. Test coverage is thorough: SDDL-shape assertions on Windows, mode assertions on Unix, atomic-rollback assertions on both, and direct regression tests for the PowerShell backslash path. The documentation in `secureperm` itself reads well and explains *why* CREATE_NEW + SECURITY_ATTRIBUTES is the load-bearing detail. A few things I'd like the author to consider before merging — the most substantive being a missing fsync on Windows that breaks the parity the package doc claims, and the `kubeconfig_handler.go` still using the bare `HasPrefix("..")` check that this PR fixed elsewhere.

## Recommendation

**Comment** — leaning approve once the Windows fsync omission and the `kubeconfig_handler.go` `HasPrefix("..")` carry-over are addressed (or explicitly punted with a follow-up issue). The rest are nits / discussion. Test coverage and the `secureperm` package docs are notably good.

---

That's the draft. Nothing has been posted; ready for your review and any edits before you send it.

pkg/secureperm/secureperm_windows.go
```diff
@@ -0,0 +1,196 @@
...
+		return "", 0, err
+	}
+	sa := windows.SecurityAttributes{
+		Length:             uint32(unsafe.Sizeof(windows.SecurityAttributes{})),
+		SecurityDescriptor: sd,
+		InheritHandle:      0,
+	}
+
+	for range 100 {
+		candidate := filepath.Join(dir, fmt.Sprintf(".secureperm-%d-%d", os.Getpid(), rand.Uint32()))
```
myasnikovdaniil, line 104: **tmp filename entropy**

```go
	for range 100 {
		candidate := filepath.Join(dir, fmt.Sprintf(".secureperm-%d-%d", os.Getpid(), rand.Uint32()))
```

`math/rand/v2` is auto-seeded per process so two concurrent talm processes in the same dir would have different streams; combined with `os.Getpid()` and the 100-attempt retry, collisions are statistically zero. Just noting the design is sound — no change requested.

---

pkg/secureperm/secureperm_windows.go
lexfrei: Done in e8c8051 — added `f.Sync()` before close so FlushFileBuffers runs on the Windows path too. Comment carries the same rationale as the Unix variant. Skipped the directory-level flush per your note that Windows has no clean public API for that; the file flush is the load-bearing call.

pkg/commands/kubeconfig_handler.go
lexfrei: Folded into this PR in 9383179. The boolean contract of `isOutsideRoot` is now also pinned directly via `TestIsOutsideRoot` in d6a069b so any future call site inherits validated semantics rather than re-deriving them.

pkg/commands/kubeconfig_handler.go
lexfrei: Same fix in 9383179.

pkg/secureperm/secureperm_windows.go
lexfrei: Dropped in e8c8051 (same commit as the fsync — both touch the same `WriteFile` entry). Agreed it read like there was a code path producing `""` and there isn't.

pkg/secureperm/secureperm_unix.go
lexfrei: Documented in b28c948. Took the no-code-change path you suggested — a paragraph in the package doc next to the atomicity claim, explaining the uid/gid difference vs `os.WriteFile`'s O_TRUNC semantics and pointing mixed-uid setups at running talm under a consistent identity.

pkg/commands/template_test.go
lexfrei: Comment added in e81053c.

pkg/commands/template_windows_test.go
lexfrei: Comment added in e81053c.

.github/workflows/pr.yml
lexfrei: Confirmed — `fail-fast: false` is intentional precisely so a Windows-specific failure surfaces even when Linux is green (and vice versa). I'll make the behavioural claim explicit in future PR descriptions rather than letting it ride on the reader recognising the strategy block.

pkg/secureperm/secureperm_windows.go
lexfrei: Thanks for the audit. Agreed — `math/rand/v2`'s per-process auto-seeding plus `os.Getpid()` plus the 100-attempt retry makes collisions a non-issue in practice.

myasnikovdaniil (dismissed):
All review comments addressed; Windows fsync-before-rename and kubeconfig_handler isOutsideRoot fixes verified. LGTM.

kvaps (approved):
LGTM. Windows support is thorough: atomic writes, proper NTFS DACL handling, comprehensive cross-platform tests, and the collateral fixes (..templates edge case, vacuous err wrapper) are appreciated.

PR #130, opened 2026-04-18
fix(chart): discovery-based defaults for subnets; required endpoint and VIP

### LINKED ISSUE #25: Bug generate template
My k8s network is 192.168.201.0/24.
Talm correctly generates interfaces, routes, name servers.
My command to generate template:
``` bash talm -n 192.168.201.201.10 -e 192.168.201.201.10 template -t templates/controlplane.yaml > nodes/kub-t10.yaml ```
Does not correctly generate validSubnets, endpoint, advertisedSubnets as shown below. Should be values from network 192.168.201.0/24

```yaml
machine:
  type: controlplane
  kubelet:
    nodeIP:
      validSubnets:
        - 192.168.100.0/24
```
and
```yaml
cluster:
  controlPlane:
    endpoint: https://192.168.100.10:6443
```
also
```yaml
  etcd:
    advertisedSubnets:
      - 192.168.100.0/24
``` 

Thank you!

lexfrei:
## Summary

Close the loop on the `192.168.100.*` placeholder problem: `talm template` on a node in any subnet other than 192.168.100.0/24 silently produced machine configs pointing at network IPs no node actually owns.

Closes #25

### 1. Subnet fields derive from discovery

`machine.kubelet.nodeIP.validSubnets` and `cluster.etcd.advertisedSubnets` now fall back to the CIDR of the node's default-gateway-bearing link when `values.yaml` leaves `advertisedSubnets` empty. The chart pipes the discovered address (`192.168.201.10/24`) through a new `cidrNetwork` template function — `net/netip`'s `ParsePrefix().Masked()` in 6 lines — so the emitted YAML is the canonical network form (`192.168.201.0/24`) rather than the host form. Multiple addresses on the same link are deduped via sprig's `uniq`.

Override precedence is preserved: if the operator sets `advertisedSubnets` explicitly in `values.yaml`, that wins.

When both paths fail (operator didn't set it, discovery returned nothing), the chart fails loudly with a `required()` error that explains both remediations (set it explicitly, or fix the node's default route) — not silently emit an empty list.

### 2. Cluster-wide fields are now `required` instead of placeholder-defaulted

Three fields shipped with `192.168.100.*` placeholder defaults that silently produced broken configs for anyone not on that subnet:

- `endpoint` — what every node's kubelet and kube-proxy dials.
- `floatingIP` (cozystack preset) — drives the Layer-2 VIP document.
- Implicitly the coupling between them: they must match for a VIP setup.

All three are now blank by default. `endpoint` triggers a `required()` guard with a verbose message that explains *why* auto-derivation isn't possible (`talm template` runs once per node and has no way to reconcile per-node IPs into a single shared endpoint) and enumerates the viable options (cozystack VIP, external LB, single-node routable IP). `floatingIP` emits no `Layer2VIPConfig` document when blank, which is the correct behavior for single-node and external-LB topologies.

### 3. Documentation matches reality

- `charts/cozystack/values.yaml` and `charts/generic/values.yaml` — rewritten comments explain the decision tree (cozystack VIP vs single-node vs external LB) and the `endpoint`/`floatingIP` coupling.
- `README.md` "Getting Started" — adds the required `values.yaml` edit step between `talm init` and `talm template`. Example values match the tutorial's node IP so a reader walking the example doesn't see placeholder-vs-tutorial drift.

### 4. Test coverage

`pkg/engine/render_test.go` gains tests for:

- `validSubnets` and `advertisedSubnets` discovery fallback (controlplane + worker + generic preset).
- Override precedence: operator `values.yaml` wins.
- Empty-discovery `required()` guard fires with both the field name and the remediation phrase in the error.
- Shipped-defaults regression guard: fresh `chrt.Values` render fails, catching any future commit that re-introduces a placeholder.
- VIP-when-set: Layer2VIPConfig emitted only when operator sets `floatingIP`.
- No-VIP on fresh defaults.
- Dedupe: two addresses in the same subnet on one link collapse to one entry per consumer.

`pkg/engine/helm/engine_test.go` gains `TestCidrNetworkTemplateFunc` — direct IPv4/IPv6 unit test of the new template function (7 cases including already-canonical, narrow prefix, malformed inputs).

## Migration notes for existing users

- Anyone who `talm init`'d before and left `values.yaml` defaults now sees a `required` error on next `talm template`. Remediation is in the error message.
- Anyone who explicitly set `endpoint` and `advertisedSubnets` in `values.yaml` is unaffected.
- Anyone who relied on the cozystack `floatingIP: 192.168.100.10` default (unlikely, given it was network-incorrect for most users) now needs to set it explicitly.

## Verification

- `go test ./...` on macOS — pass
- `GOOS=windows GOARCH=amd64 go build ./...` — clean
- `golangci-lint run ./...` — 0 issues

## Release Notes

* **Documentation**
  * Enhanced README with detailed configuration guidance for control-plane endpoint and VIP setup across deployment scenarios.

* **Configuration Changes**
  * Removed default values for control-plane endpoint, floating IP, and subnet configuration; explicit values now required.

* **Improvements**
  * Added automatic subnet discovery from network interfaces when not explicitly configured.
  * Strengthened validation with clearer error messages for required settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,22 +72,22 @@ automatically from the node's default-gateway-bearing link, so no
 override is needed unless you have a multi-homed node that requires
 a specific subnet pinned.
 
-Boot Talos Linux node, let's say it has address `1.2.3.4`. Then:
+Boot Talos Linux node, let's say it has address `192.0.2.4`. Then:
 
 ```yaml
-# values.yaml (single-node example matching the 1.2.3.4 node below)
-endpoint: "https://1.2.3.4:6443"
+# values.yaml (single-node example matching the 192.0.2.4 node below)
+endpoint: "https://192.0.2.4:6443"
 floatingIP: ""
 ```
 
 Gather node information:
 ```bash
-talm -n 1.2.3.4 -e 1.2.3.4 template -t templates/controlplane.yaml -i > nodes/node1.yaml
+talm -n 192.0.2.4 -e 192.0.2.4 template -t templates/controlplane.yaml -i > nodes/node1.yaml
 ```
 
 Edit `nodes/node1.yaml` file:
 ```yaml
-# talm: nodes=["1.2.3.4"], endpoints=["1.2.3.4"], templates=["templates/controlplane.yaml"]
+# talm: nodes=["192.0.2.4"], endpoints=["192.0.2.4"], templates=["templates/controlplane.yaml"]
 machine:
     network:
         # -- Discovered interfaces:
@@ -108,10 +108,10 @@ machine:
         interfaces:
             - interface: enx9c6b0047066c
               addresses:
-                - 1.2.3.4/26
+                - 192.0.2.4/26
               routes:
                 - network: 0.0.0.0/0
-                  gateway: 1.2.3.1
+                  gateway: 192.0.2.1
         nameservers:
             - 8.8.8.8
             - 8.8.4.4
@@ -132,7 +132,7 @@ machine:
 cluster:
     clusterName: talm
     controlPlane:
-        endpoint: https://192.168.0.1:6443
+        endpoint: https://192.0.2.4:6443
 ```
 
 > **Note:** The output format depends on the Talos version configured in `Chart.yaml` (`templateOptions.talosVersion`) or via the `--talos-version` CLI flag.
--- a/README.md
+++ b/README.md
@@ -59,6 +59,24 @@ cd newcluster
 talm init -p cozystack -N myawesomecluster
 ```
 
+Edit `values.yaml` to set your cluster's control-plane endpoint — the
+URL every node's kubelet and kube-proxy will dial. The chart leaves
+it empty on purpose so a missed override fails loudly instead of
+silently embedding a placeholder. For cozystack deployments set
+`endpoint` and `floatingIP` together (same IP, single VIP); for
+single-node clusters use that node's routable IP; for multi-node
+with an external load balancer use the LB URL. Subnet-selector
+fields (`kubelet.validSubnets`, `etcd.advertisedSubnets`) are derived
+automatically from the node's default-gateway-bearing link — no
+override needed unless you have a multi-homed node that requires a
+specific subnet pinned.
+
+```yaml
+# values.yaml
+endpoint: "https://192.168.0.1:6443"
+floatingIP: 192.168.0.1
+```
+
 Boot Talos Linux node, let's say it has address `1.2.3.4`
 
 Gather node information:
--- a/README.md
+++ b/README.md
@@ -59,25 +59,26 @@ cd newcluster
 talm init -p cozystack -N myawesomecluster
 ```
 
-Edit `values.yaml` to set your cluster's control-plane endpoint — the
-URL every node's kubelet and kube-proxy will dial. The chart leaves
-it empty on purpose so a missed override fails loudly instead of
-silently embedding a placeholder. For cozystack deployments set
-`endpoint` and `floatingIP` together (same IP, single VIP); for
-single-node clusters use that node's routable IP; for multi-node
-with an external load balancer use the LB URL. Subnet-selector
-fields (`kubelet.validSubnets`, `etcd.advertisedSubnets`) are derived
-automatically from the node's default-gateway-bearing link — no
-override needed unless you have a multi-homed node that requires a
-specific subnet pinned.
+Edit `values.yaml` to set your cluster's control-plane endpoint. This
+is the URL every node's kubelet and kube-proxy will dial. The chart
+leaves it empty on purpose so a missed override fails loudly instead
+of silently embedding a placeholder. For cozystack VIP setups set
+`endpoint` and `floatingIP` together (same IP, single shared VIP);
+for single-node clusters use that node's routable IP and leave
+`floatingIP` blank; for multi-node with an external load balancer
+use the LB URL and leave `floatingIP` blank. Subnet-selector fields
+(`kubelet.validSubnets`, `etcd.advertisedSubnets`) are derived
+automatically from the node's default-gateway-bearing link, so no
+override is needed unless you have a multi-homed node that requires
+a specific subnet pinned.
+
+Boot Talos Linux node, let's say it has address `1.2.3.4`. Then:
 
 ```yaml
-# values.yaml
-endpoint: "https://192.168.0.1:6443"
-floatingIP: 192.168.0.1
+# values.yaml (single-node example matching the 1.2.3.4 node below)
+endpoint: "https://1.2.3.4:6443"
+floatingIP: ""
 ```
-
-Boot Talos Linux node, let's say it has address `1.2.3.4`
 
 Gather node information:
 ```bash
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -18,7 +18,14 @@ machine:
   kubelet:
     nodeIP:
       validSubnets:
+        {{- if .Values.advertisedSubnets }}
         {{- toYaml .Values.advertisedSubnets | nindent 8 }}
+        {{- else }}
+        {{- /* Fall back to the CIDR of the node's default-gateway-bearing link. */ -}}
+        {{- range fromJsonArray (include "talm.discovered.default_addresses_by_gateway" .) }}
+        - {{ . }}
+        {{- end }}
+        {{- end }}
     extraConfig:
       cpuManagerPolicy: static
       maxPods: 512
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -21,15 +21,22 @@ machine:
         {{- if .Values.advertisedSubnets }}
         {{- toYaml .Values.advertisedSubnets | nindent 8 }}
         {{- else }}
-        {{- /* Fall back to the subnet of the node's default-gateway-bearing link.
-               cidrNetwork masks host bits so the emitted YAML is the canonical
-               network form (192.168.201.0/24) rather than host-form (192.168.201.10/24). */ -}}
+        {{- /* Fall back to the subnet of the node's default-gateway-bearing
+               link. cidrNetwork masks host bits so the emitted YAML is the
+               canonical network form (192.168.201.0/24) rather than the
+               host form (192.168.201.10/24). Dedupe after masking because
+               a link with a secondary address in the same subnet would
+               otherwise produce duplicate list entries. */ -}}
         {{- $addrs := fromJsonArray (include "talm.discovered.default_addresses_by_gateway" .) }}
         {{- if not $addrs }}
-        {{- required "values.yaml: `advertisedSubnets` was left empty and talm could not derive a default from discovery — no default-gateway-bearing link was found on the node. Either set advertisedSubnets explicitly in values.yaml, or ensure the node has a default route before running `talm template`." "" }}
+        {{- required "values.yaml: `advertisedSubnets` was left empty and talm could not derive a default from discovery. No default-gateway-bearing link was found on the node. This field is a cluster-wide subnet selector fed to kubelet and etcd; `talm template` is invoked once per node and cannot merge per-node values into one cluster value. Either set advertisedSubnets explicitly in values.yaml, or ensure the node has a default route before running `talm template`." "" }}
         {{- end }}
+        {{- $subnets := list }}
         {{- range $addrs }}
-        - {{ . | cidrNetwork }}
+        {{- $subnets = append $subnets (. | cidrNetwork) }}
+        {{- end }}
+        {{- range uniq $subnets }}
+        - {{ . }}
         {{- end }}
         {{- end }}
     extraConfig:
@@ -98,7 +105,7 @@ cluster:
       {{- toYaml .Values.serviceSubnets | nindent 6 }}
   clusterName: "{{ .Chart.Name }}"
   controlPlane:
-    endpoint: {{ required "values.yaml: `endpoint` must be set to the cluster control-plane endpoint (e.g. https://<vip>:6443); no auto-discovery is possible because this is a cluster-wide value, not a per-node one" .Values.endpoint | quote }}
+    endpoint: {{ required "values.yaml: `endpoint` must be set to the cluster control-plane URL (e.g. https://<vip>:6443). This field is cluster-wide: every node's kubelet and kube-proxy dials it, so it cannot be auto-derived from the current node's IP -- `talm template` runs once per node and has no way to reconcile per-node IPs into a single shared endpoint. For multi-node setups use a VIP (cozystack floatingIP) or an external load balancer; for single-node clusters the node's routable IP works." .Values.endpoint | quote }}
   {{- if eq .MachineType "controlplane" }}
   allowSchedulingOnControlPlanes: true
   controllerManager:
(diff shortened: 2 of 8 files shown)


charts/cozystack/templates/_helpers.tpl
```diff
@@ -18,7 +18,27 @@ machine:
...
+        {{- else }}
+        {{- /* Fall back to the subnet of the node's default-gateway-bearing
+               link. cidrNetwork masks host bits so the emitted YAML is the
+               canonical network form (192.168.201.0/24) rather than the
+               host form (192.168.201.10/24). Dedupe after masking because
+               a link with a secondary address in the same subnet would
+               otherwise produce duplicate list entries. */ -}}
+        {{- $addrs := fromJsonArray (include "talm.discovered.default_addresses_by_gateway" .) }}
+        {{- if not $addrs }}
+        {{- required "values.yaml: `advertisedSubnets` was left empty and talm could not derive a default from discovery. No default-gateway-bearing link was found on the node. This field is a cluster-wide subnet selector fed to kubelet and etcd; `talm template` is invoked once per node and cannot merge per-node values into one cluster value. Either set advertisedSubnets explicitly in values.yaml, or ensure the node has a default route before running `talm template`." "" }}
```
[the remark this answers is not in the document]
lexfrei: Good call — swapped to `fail` in 42967b7. `required ""` was a hack to express unconditional failure, `fail` says the same thing directly and matches how the Helm engine already registers both helpers side by side.

charts/generic/templates/_helpers.tpl
```diff
@@ -13,7 +13,27 @@ machine:
...
+        {{- else }}
+        {{- /* Fall back to the subnet of the node's default-gateway-bearing
+               link. cidrNetwork masks host bits so the emitted YAML is the
+               canonical network form (192.168.201.0/24) rather than the
+               host form (192.168.201.10/24). Dedupe after masking because
+               a link with a secondary address in the same subnet would
+               otherwise produce duplicate list entries. */ -}}
+        {{- $addrs := fromJsonArray (include "talm.discovered.default_addresses_by_gateway" .) }}
+        {{- if not $addrs }}
+        {{- required "values.yaml: `advertisedSubnets` was left empty and talm could not derive a default from discovery. No default-gateway-bearing link was found on the node. This field is a cluster-wide subnet selector fed to kubelet and etcd; `talm template` is invoked once per node and cannot merge per-node values into one cluster value. Either set advertisedSubnets explicitly in values.yaml, or ensure the node has a default route before running `talm template`." "" }}
```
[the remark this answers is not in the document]
lexfrei: Fixed symmetrically in 3a8c0a5 — the generic preset now uses `fail` too, keeping the two presets identical.

pkg/engine/render_test.go
```diff
@@ -48,8 +62,22 @@ func renderChartTemplate(t *testing.T, chartPath string, templateFile string, ta
 		tv = talosVersion[0]
 	}
 
+	// Inject test defaults when the chart ships empty values (true
+	// for cozystack and generic presets post-issue-25 fix). Tests
+	// that specifically exercise the required-endpoint or empty-
+	// discovery guards build their own values maps and do not go
+	// through this helper.
+	values := make(map[string]any)
+	maps.Copy(values, chrt.Values)
```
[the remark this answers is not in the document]
lexfrei: Fair point for future-proofing — switched in 6295e02. Added a small `cloneValues` helper that recursively deep-copies maps and slices (primitives are immutable so they can stay shared) and routed `renderCozystackWith`, `renderGenericWith`, and `renderChartTemplate` through it. The remaining `maps.Copy` is only for merging caller-supplied overrides, and those maps are built fresh per test so they're already safe.

pkg/engine/render_test.go
```diff
@@ -964,3 +1048,482 @@ func TestRenderInvalidTalosVersion(t *testing.T) {
...
+				}
+				if id == "" {
+					return map[string]any{"apiVersion": "v1", "kind": "List", "items": []any{eth0}}, nil
+				}
+			case "addresses":
+				return addressesList, nil
+			case "nodeaddress":
+				if id == "default" {
+					return map[string]any{"spec": map[string]any{"addresses": []any{"192.168.201.10/24"}}}, nil
+				}
```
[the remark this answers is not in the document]
lexfrei: The dedupe path here runs through `talm.discovered.default_addresses_by_gateway`, which reads the `addresses` resource (see `charts/talm/templates/_helpers.tpl:75`) — not `nodeaddress/default`. The `addresses` fixture already publishes two entries on eth0 in the same /24, so the duplicate input is exercised as intended. `nodeaddress/default` feeds `talm.discovered.default_addresses`, a separate helper that is not on the subnet-derivation path, so the single-address entry there does not weaken the regression test.

README.md
```diff
@@ -59,7 +59,26 @@ cd newcluster
...
+override is needed unless you have a multi-homed node that requires
+a specific subnet pinned.
+
+Boot Talos Linux node, let's say it has address `1.2.3.4`. Then:
+
+```yaml
+# values.yaml (single-node example matching the 1.2.3.4 node below)
+endpoint: "https://1.2.3.4:6443"
+floatingIP: ""
+```
```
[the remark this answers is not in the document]
lexfrei: Addressed in d67d0d1e — swapped the whole getting-started walkthrough (values.yaml example, CLI invocation, rendered config sample) to the RFC 5737 TEST-NET-1 range (192.0.2.0/24) so the documentation uses IPs reserved for examples. The rendered endpoint now reads `https://192.0.2.4:6443`, matching the values.yaml example above.

charts/cozystack/templates/_helpers.tpl
```diff
@@ -119,7 +139,23 @@ cluster:
     enabled: false
   etcd:
     advertisedSubnets:
+      {{- if .Values.advertisedSubnets }}
       {{- toYaml .Values.advertisedSubnets | nindent 6 }}
+      {{- else }}
+      {{- /* Fall back to the subnet of the node's default-gateway-bearing
+             link; cidrNetwork masks host bits to emit canonical network
+             form. Dedupe handled the same way as validSubnets above.
+             Empty discovery already errored via validSubnets' required()
```
myasnikovdaniil, line 148: **stale "required()" wording in comment**

The most recent commit (`3a8c0a5`) switched from `required` to `fail` for the empty-discovery path, but the comment in the etcd block still references the old guard name. Same wording lives in `charts/generic/templates/_helpers.tpl:71`.

```tpl
      {{- /* Fall back to the subnet of the node's default-gateway-bearing
             link; cidrNetwork masks host bits to emit canonical network
             form. Dedupe handled the same way as validSubnets above.
             Empty discovery already errored via validSubnets' fail
             guard, so we reach this block only when at least one address
             was resolved. */ -}}
```

charts/cozystack/templates/_helpers.tpl
```diff
@@ -119,7 +139,23 @@ cluster:
...
+             guard, so we reach this block only when at least one address
+             was resolved. */ -}}
+      {{- $subnets := list }}
+      {{- range fromJsonArray (include "talm.discovered.default_addresses_by_gateway" .) }}
+      {{- $subnets = append $subnets (. | cidrNetwork) }}
+      {{- end }}
+      {{- range uniq $subnets }}
+      - {{ . }}
+      {{- end }}
+      {{- end }}
```
myasnikovdaniil, line 158: **etcd block's empty-discovery handling is implicit**

The etcd `advertisedSubnets` block deliberately omits the empty-discovery `fail` because validSubnets renders first and errors out earlier. That's true today, but fragile: any future refactor that splits `talos.config.machine.common` and `talos.config.cluster` into different render passes (or skips the machine block for some path) silently emits an empty `advertisedSubnets:` list — exactly the silent-fail mode this PR is trying to eliminate.

A defensive duplicate of the same `if not $addrs / fail` check costs five lines and removes the cross-block dependency. Same applies to `charts/generic/templates/_helpers.tpl:65-82`.

```tpl
      {{- else }}
      {{- $addrs := fromJsonArray (include "talm.discovered.default_addresses_by_gateway" .) }}
      {{- if not $addrs }}
      {{- fail "values.yaml: `advertisedSubnets` was left empty and talm could not derive a default from discovery. ..." }}
      {{- end }}
      {{- $subnets := list }}
      {{- range $addrs }}
      {{- $subnets = append $subnets (. | cidrNetwork) }}
      {{- end }}
      {{- range uniq $subnets }}
      - {{ . }}
      {{- end }}
      {{- end }}
```

charts/cozystack/templates/_helpers.tpl
```diff
@@ -18,7 +18,27 @@ machine:
   kubelet:
     nodeIP:
       validSubnets:
+        {{- if .Values.advertisedSubnets }}
         {{- toYaml .Values.advertisedSubnets | nindent 8 }}
```
myasnikovdaniil, line 22: **operator-override path skips canonicalization**

When the operator sets `advertisedSubnets: ["192.168.201.10/24"]` (host form) in values.yaml, the chart emits it verbatim via `toYaml`, while the fallback path runs every entry through `cidrNetwork` so the YAML is in canonical network form. That inconsistency means two operators with the same underlying intent — one relying on discovery, one with an explicit override — get differently-formatted machine configs, which is noisy in `kubectl diff`/`talm diff` output and surprising in code review. Consider canonicalizing in both branches:

```tpl
      validSubnets:
        {{- $sources := .Values.advertisedSubnets }}
        {{- if not $sources }}
        {{- $sources = fromJsonArray (include "talm.discovered.default_addresses_by_gateway" .) }}
        {{- if not $sources }}
        {{- fail "values.yaml: `advertisedSubnets` was left empty and talm could not derive a default from discovery. ..." }}
        {{- end }}
        {{- end }}
        {{- $subnets := list }}
        {{- range $sources }}
        {{- $subnets = append $subnets (. | cidrNetwork) }}
        {{- end }}
        {{- range uniq $subnets }}
        - {{ . }}
        {{- end }}
```

This also de-duplicates the validSubnets and etcd.advertisedSubnets blocks down to a shared `talos.discovered.subnet_list` define, which is a sympathetic refactor for the comment in the etcd block that already says "handled the same way as validSubnets above." If you'd rather keep the operator-override path passthrough (e.g., to preserve operator-supplied formatting verbatim), call this out in a comment so future readers don't try to "fix" the inconsistency.

charts/cozystack/templates/_helpers.tpl
```diff
@@ -18,7 +18,27 @@ machine:
...
+        {{- else }}
+        {{- /* Fall back to the subnet of the node's default-gateway-bearing
+               link. cidrNetwork masks host bits so the emitted YAML is the
+               canonical network form (192.168.201.0/24) rather than the
+               host form (192.168.201.10/24). Dedupe after masking because
+               a link with a secondary address in the same subnet would
+               otherwise produce duplicate list entries. */ -}}
+        {{- $addrs := fromJsonArray (include "talm.discovered.default_addresses_by_gateway" .) }}
+        {{- if not $addrs }}
+        {{- fail "values.yaml: `advertisedSubnets` was left empty and talm could not derive a default from discovery. No default-gateway-bearing link was found on the node. This field is a cluster-wide subnet selector fed to kubelet and etcd; `talm template` is invoked once per node and cannot merge per-node values into one cluster value. Either set advertisedSubnets explicitly in values.yaml, or ensure the node has a default route before running `talm template`." }}
```
myasnikovdaniil, line 32: **fail message field-name disambiguation**

The error fires from inside the `validSubnets:` rendering block but the message references `advertisedSubnets`. The values.yaml key is `advertisedSubnets`, so the message *is* correct, but a user staring at the rendered (broken) output sees `validSubnets:` and may chase the wrong field. One word in the message would close the gap:

```
"values.yaml: `advertisedSubnets` (which feeds both kubelet validSubnets and etcd advertisedSubnets) was left empty and talm could not derive a default from discovery. ..."
```

Same in the generic copy.

pkg/engine/helm/engine.go
```diff
@@ -218,6 +219,19 @@ func (e Engine) initFunMap(t *template.Template) {
...
+	// cidrNetwork canonicalizes a CIDR string to its network form
+	// ("192.168.201.10/24" -> "192.168.201.0/24"), matching what
+	// operators see in Talos docs and upstream examples. Sprig ships
+	// no equivalent; net/netip's ParsePrefix + Masked handles both
+	// IPv4 and IPv6 without any host-bit arithmetic in the template.
+	funcMap["cidrNetwork"] = func(cidr string) (string, error) {
+		p, err := netip.ParsePrefix(cidr)
+		if err != nil {
+			return "", fmt.Errorf("cidrNetwork: %w", err)
+		}
```
myasnikovdaniil, line 231: **`cidrNetwork` is fine; consider also exposing the bare network address**

Not a request, just a flag for future iterations: callers may eventually want the network address without the prefix length (`192.168.201.0` rather than `192.168.201.0/24`) — for example when constructing CIDR list entries with explicit prefix logic. If a `cidrNetworkAddress` companion is ever needed, the `netip.Prefix.Masked().Addr().String()` form is one line. Out of scope for this PR.

charts/cozystack/values.yaml
```diff
@@ -1,13 +1,53 @@
...
-floatingIP: 192.168.100.10
+# Layer-2 VIP for cozystack multi-node setups. When set, the chart
+# emits a Layer2VIPConfig document pinning this IP as a floating
+# address on the node's primary link. MUST equal the host portion
+# of `endpoint` above, otherwise the cluster dials an IP that no
+# node actually claims. Blank by default so the shipped value never
+# silently embeds a wrong VIP -- fill in only if you want a VIP.
+# Single-node clusters and external-LB topologies leave it blank.
+# Example: floatingIP: 192.168.0.1
+floatingIP: ""
```
myasnikovdaniil, line 31: **coupling note in `floatingIP` comment is great, suggest explicit cross-link**

The `floatingIP` comment says "MUST equal the host portion of `endpoint` above" but doesn't show the symmetric note in the `endpoint` comment (it does mention "set `endpoint` AND `floatingIP` below to the SAME IP"). The two comments together tell the right story, but a reader who hits one comment first without scrolling may miss the coupling. Consider a one-line cross-reference at the top of `floatingIP`:

```yaml
# See `endpoint` above — these two values are coupled and MUST match
# for cozystack VIP setups.
```

myasnikovdaniil (commented):
Here is my draft review.

---

# Draft review for PR #130 — fix(chart): discovery-based defaults for subnets; required endpoint and VIP

## Summary

This PR closes a real foot-gun: charts shipping with `192.168.100.*` placeholder defaults silently produced broken machine configs for everyone not on that subnet. The fix is in three layers — empty-by-default required values for cluster-wide fields (`endpoint`, `floatingIP`), a discovery fallback wired through a new `cidrNetwork` template function for `advertisedSubnets`, and rewritten user-facing docs that explain the cozystack-VIP / single-node / external-LB decision tree. Test coverage is thorough (override precedence, empty-discovery guard, dedupe, fresh-defaults regression guard, IPv4 + IPv6 cidrNetwork unit tests, worker path, generic path). The breaking change for existing users is acknowledged and the error messages enumerate remediation, so the migration story is reasonable. A few minor cleanups noted below; nothing blocking.

### Migration safety check — schema validation as a nice-to-have

This isn't a blocker, but since `endpoint` is now hard-required and the failure mode is "render-time error with a long message," adding a `values.schema.json` to each chart that marks `endpoint` as required would surface the issue at `helm lint` time too — earlier in the workflow than a render error. Out of scope for this PR; flagging only because the PR is the natural place to do it next.

## Recommendation

**Comment** — the change is sound and well-tested. The findings above are nits and small defensive-programming suggestions; none of them block merge. I'd particularly encourage fixing the stale `required()`-vs-`fail` comment (5-second change) and adding the duplicate empty-discovery guard to the etcd block (5-line change) before merge. The CIDR canonicalization consistency and the schema-validation suggestion can land later.

---

Stopping here — this is a draft for your review, not posted anywhere.

### `cloneValues` / `deepClone` justification reads well (`pkg/engine/render_test.go`)

The deep-copy helper is well-justified by the comment ("maps.Copy is a shallow copy — mutating a nested map or slice in a test would leak into chrt.Values and corrupt subsequent renders"). Nice.

kvaps (approved):

PR #131, opened 2026-04-27
feat(cozystack): add DRBD-oriented sysctl and etcd backend defaults

### LINKED ISSUE #2332: feat: add recommended sysctl and etcd defaults to Talos configuration
Add recommended sysctl and etcd defaults to the Talos machine configuration.

## Proposed defaults

```yaml
machine:
  sysctls:
    # TCP orphan handling
    net.ipv4.tcp_orphan_retries: "3"
    net.ipv4.tcp_fin_timeout: "30"

    # Network backlog
    net.core.netdev_max_backlog: "5000"
    net.core.netdev_budget: "600"
    net.core.netdev_budget_usecs: "8000"

    # TCP keepalive (early detection of dead connections)
    net.ipv4.tcp_keepalive_time: "600"
    net.ipv4.tcp_keepalive_intvl: "10"
    net.ipv4.tcp_keepalive_probes: "6"

  etcd:
    extraArgs:
      quota-backend-bytes: "8589934592"
      max-request-bytes: "10485760"
```

## Why

- **tcp_orphan_retries=3**: Talos default is 0, meaning a single lost ACK causes orphaned TCP connections to accumulate indefinitely. With DRBD creating many peer connections, this leads to port exhaustion.
- **tcp_fin_timeout=30**: Reduces TIME_WAIT duration for faster port recycling.
- **netdev_max_backlog/budget**: Prevents packet loss under high network load from DRBD replication traffic.
- **tcp_keepalive**: Enables faster detection of dead DRBD peer connections (10min idle + 60s probe window instead of default 2h+75s).
- **etcd quota-backend-bytes=8Gi**: Prevents etcd running out of space with large LINSTOR CRD datasets.
- **etcd max-request-bytes=10Mi**: Allows larger CRD objects to be stored.

IvanHunters:
## Summary

Adds DRBD-oriented sysctl and etcd backend defaults to the `cozystack` preset. Cozystack nodes always run DRBD; reconnect storms (node reboots, resync) exhaust TCP ports under the default Talos sysctl profile, and a LINSTOR-heavy control plane can outgrow etcd's default 2 GiB backend.

### Always-on `machine.sysctls`

- `tcp_orphan_retries=3`, `tcp_fin_timeout=30` — reclaim orphaned and FIN-WAIT sockets faster so a reconnect storm cannot outrun cleanup.
- `netdev_max_backlog=5000`, `netdev_budget=600`, `netdev_budget_usecs=8000` — widen the receive backlog so bursty replication traffic isn't dropped under load.

These five are the tuning that resolved the TCP-port exhaustion on production clusters (see siderolabs/talos#13074).

### Opt-in TCP keepalive (`tcpKeepaliveTuning`, default off)

`tcp_keepalive_time=600` / `intvl=10` / `probes=6` is gated behind a new `tcpKeepaliveTuning` value, off by default. These sysctls are kernel-wide — they shorten idle-socket failure detection for every long-lived TCP connection on the node (NFS mounts, DB clients, MQ consumers), not just DRBD. DRBD already detects dead peers in seconds via its own protocol-level ping (`ping-int`/`ping-timeout`), so this is a generic socket backstop rather than a DRBD requirement; hence opt-in rather than baked in.

### Tunable etcd backend quota (`etcd.quotaBackendBytes`, default 8 GiB)

`cluster.etcd.extraArgs.quota-backend-bytes` is exposed via a new `etcd.quotaBackendBytes` value (default 8 GiB, etcd's documented upper bound), emitted on controlplane nodes only. It raises etcd's 2 GiB ceiling so a control plane holding many DRBD-resource CRDs in aggregate does not trip the NOSPACE alarm. It is a ceiling, not a reservation — a small DB stays small and costs no extra RAM/disk — and can be blanked to fall back to etcd's own default.

### Dropped from the original proposal: etcd `max-request-bytes`

The source issue also proposed `max-request-bytes=10MiB`. On verification this does not achieve its stated goal: LINSTOR k8s-backend writes go through kube-apiserver, whose `MaxRequestBodyBytes` is hardcoded at 3 MiB with no configuration flag (kubernetes/kubernetes#88968 — "No flag was intended for this option"). The effective per-object ceiling stays at the apiserver's 3 MiB regardless of etcd's setting, and the real "large dataset" concern is aggregate DB size, which `quota-backend-bytes` already covers. The knob was dropped rather than shipped as a no-op.

## Tests

- New contract tests pin every branch: the always-on DRBD sysctls (present on cozystack, absent on generic), the keepalive toggle (absent by default, present when enabled, operator-settable while off, collision-protected while on), and the etcd quota (default, tunable, omitted-when-blank, controlplane-only, absent on generic).
- `docs/manual-test-plan.md` gains a scenario (B9) exercising each of the above.

## Docs

Operator-facing documentation updated in a companion PR: cozystack/website#567.

## Related

- Resolves cozystack/cozystack#2332
- Upstream discussion: siderolabs/talos#13074

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in TCP keepalive tuning toggle and a configurable etcd backend quota (raised default) for improved network reliability and reduced NOSPACE risk.
* **Chores**
  * Broadened preset sysctl tuning for network/DRBD performance (non-overridable by operators when preset active).
* **Documentation**
  * Expanded manual test plan with validation steps for sysctl and etcd quota behavior.
* **Tests**
  * Added contract and render tests covering sysctl behavior and etcd quota rendering.

### DIFF
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -29,6 +29,17 @@ machine:
     net.ipv4.neigh.default.gc_thresh1: "4096"
     net.ipv4.neigh.default.gc_thresh2: "8192"
     net.ipv4.neigh.default.gc_thresh3: "16384"
+    # TCP orphan handling
+    net.ipv4.tcp_orphan_retries: "3"
+    net.ipv4.tcp_fin_timeout: "30"
+    # Network backlog
+    net.core.netdev_max_backlog: "5000"
+    net.core.netdev_budget: "600"
+    net.core.netdev_budget_usecs: "8000"
+    # TCP keepalive (early detection of dead connections)
+    net.ipv4.tcp_keepalive_time: "600"
+    net.ipv4.tcp_keepalive_intvl: "10"
+    net.ipv4.tcp_keepalive_probes: "6"
   kernel:
     modules:
     - name: openvswitch
@@ -120,6 +131,9 @@ cluster:
   etcd:
     advertisedSubnets:
       {{- toYaml .Values.advertisedSubnets | nindent 6 }}
+    extraArgs:
+      quota-backend-bytes: "8589934592"  # 8GiB - prevent etcd running out of space with large LINSTOR CRD datasets
+      max-request-bytes: "10485760"      # 10MiB - allow larger CRD objects to be stored
   {{- end }}
 {{- end }}
 
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -58,9 +58,43 @@ machine:
       {{- toYaml . | nindent 6 }}
       {{- end }}
   {{- /* extraSysctls MUST NOT collide with the preset's built-in
-         sysctls; same rationale as extraKubeletExtraArgs. */ -}}
+         sysctls; same rationale as extraKubeletExtraArgs. $builtinSysctls
+         is the single source of truth for the preset-owned keys — keep
+         it in sync with the literal sysctls block rendered further down.
+
+         Always-on DRBD/LINSTOR tuning: Cozystack always runs DRBD (the
+         drbd module is loaded unconditionally below), and these knobs
+         resolve the TCP-port exhaustion the Cozystack team observed on
+         production clusters under DRBD reconnect storms (node reboots,
+         resync). tcp_orphan_retries/tcp_fin_timeout speed up reclamation
+         of orphaned and FIN-WAIT sockets so a reconnect storm cannot
+         outrun cleanup; netdev_* widen the receive backlog so bursty
+         replication traffic isn't dropped under load.
+
+         vm.nr_hugepages is treated as preset-owned even when its gate
+         (.Values.nr_hugepages) is inactive, so operators always route it
+         through the dedicated `nr_hugepages` key. The tcp_keepalive_*
+         triplet is preset-owned only while .Values.tcpKeepaliveTuning is
+         set (see below), so it can be operator-supplied via extraSysctls
+         when the toggle is off. */ -}}
+  {{- $builtinSysctls := list
+        "vm.nr_hugepages"
+        "net.ipv4.neigh.default.gc_thresh1"
+        "net.ipv4.neigh.default.gc_thresh2"
+        "net.ipv4.neigh.default.gc_thresh3"
+        "net.ipv4.tcp_orphan_retries"
+        "net.ipv4.tcp_fin_timeout"
+        "net.core.netdev_max_backlog"
+        "net.core.netdev_budget"
+        "net.core.netdev_budget_usecs" }}
+  {{- if $.Values.tcpKeepaliveTuning }}
+  {{- $builtinSysctls = concat $builtinSysctls (list
+        "net.ipv4.tcp_keepalive_time"
+        "net.ipv4.tcp_keepalive_intvl"
+        "net.ipv4.tcp_keepalive_probes") }}
+  {{- end }}
   {{- range $k, $_ := .Values.extraSysctls }}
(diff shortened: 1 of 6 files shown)


kvaps (approved):

lexfrei (changes requested):
### Title scope

Recent PRs touching `charts/cozystack/templates/_helpers.tpl` use a concept-style scope rather than a path: `feat(cozystack): enable allocateNodeCIDRs by default` (#91), `fix(chart): cast metadata.id to string for regexMatch …` (#110), `feat(helpers): add bond interface discovery helpers` (#94). `feat(charts/cozystack): …` is the first path-style scope I see in this repo. `feat(cozystack):` would mirror #91, which was a similar defaults-tweak for this same chart.

### Merge ordering with #130

This PR edits `cluster.etcd.advertisedSubnets` immediately adjacent to where #130 wraps the same block in an `if/else` discovery fallback. There's no semantic conflict, but `git merge-tree` reports a textual conflict — whichever PR lands second will need a small rebase.

### TCP keepalive blast radius

`tcp_keepalive_time/intvl/probes = 600/10/6` is roughly 12× more aggressive on idle timeout than the kernel defaults (`7200/75/9`). The PR body frames this as DRBD-focused, but `net.ipv4.tcp_keepalive_*` are kernel-wide and apply to **every** long-lived idle TCP socket on the node — pods holding NFS mounts, idle DB clients, MQ consumers, etc. all inherit the new failure characteristics. Likely still the right default for a Cozystack cluster, but worth calling out explicitly in the PR description so operators understand the scope.

The other sysctls (`tcp_orphan_retries`, `tcp_fin_timeout`, `netdev_*`) are more conservative and look fine.

### etcd quota and request size

`quota-backend-bytes: 8 GiB` is etcd's documented **upper recommended ceiling**, not a typical default — etcd's own default is 2 GiB. Same direction for `max-request-bytes: 10 MiB` (vs etcd's 1.5 MiB default). Both are correct numbers for a LINSTOR-heavy control plane, but they're strong opinions to bake in for every Cozystack cluster: a small dev cluster still pays for the larger backend in etcd RSS regardless of whether it ever uses it.

Could these be exposed in `values.yaml` with the proposed numbers as defaults? For example:

```yaml
etcd:
  quotaBackendBytes: "8589934592"
  maxRequestBytes:   "10485760"
```

The recommendation still ships as the default, but operators can tune without forking the chart. #91 used this exact pattern for `allocateNodeCIDRs` — defaults-tweak plus values exposure in the same change.

### Tests

`pkg/engine/render_test.go` already asserts on the rendered `sysctls:` block (lines 114, 211). It would be cheap to extend those assertions to the new sysctl keys, and to add a check that `etcd.extraArgs` renders only for `MachineType: controlplane` — that's already correct in the template (the block sits inside `{{- if eq .MachineType "controlplane" }}`), but it isn't covered by a test today.

### Style nit

Inline YAML comments after string scalars (`quota-backend-bytes: "8589934592"  # 8GiB - …`) parse fine, but the rest of this template uses Go template comments (`{{- /* … */ -}}`) for explanations — see #130 for examples. Aligning would help readability.

lexfrei (approved):
LGTM. Reworked on top of current main and addressed the review points:

- Scope renamed to `feat(cozystack):`.
- etcd quota exposed as a tunable value (`etcd.quotaBackendBytes`, default 8 GiB) instead of a hardcoded arg; emitted on controlplane nodes only.
- TCP keepalive moved behind an opt-in `tcpKeepaliveTuning` (default off): it is kernel-wide, and DRBD detects dead peers in seconds via its own protocol-level ping, so it is a generic socket backstop rather than a DRBD requirement.
- The five DRBD/netdev sysctls stay always-on (low blast radius, production-validated).
- etcd `max-request-bytes` was dropped: kube-apiserver's 3 MiB request-body limit has no configuration flag (kubernetes/kubernetes#88968), so the etcd bump could not achieve its stated goal; the aggregate-size concern is covered by the quota.
- Contract tests cover every branch; the manual test plan gains a matching scenario; inline YAML comments replaced with Go-template comments.

Operator-facing docs: cozystack/website#567.

PR #147, opened 2026-05-07
fix(charts): rewrite v1.12 multi-doc network renderer for full link coverage

### LINKED ISSUE #140: bug(chart): VLAN sub-interface missing from talm template output on v1.12 multi-doc path
## What

A VLAN sub-interface that is configured on a Talos node is not emitted as a `VLANConfig` document by `talm template` on the v1.12 multi-doc rendering path. The node ends up with the VLAN silently dropped from its rendered config.

## Repro

Node has a VLAN sub-interface configured (carrying the default route, in the reporter's setup). Then:

```bash
talm template -t templates/controlplane.yaml -n <NODE_IP> -e <NODE_IP> > nodes/cozy-01.yaml
```

Expected: `nodes/cozy-01.yaml` contains an `apiVersion: v1alpha1, kind: VLANConfig` document describing the VLAN.

Observed: no `VLANConfig` in the output. The VLAN information is absent.

## Affected versions

Regression introduced by the migration to the Talos v1.12 multi-doc config format in #116 (v0.24.0). The legacy v1alpha1 single-doc render path is not affected.

## Candidate root causes (need confirmation against the live repro)

The multi-doc renderer in `charts/generic/templates/_helpers.tpl` (`talos.config.network.multidoc`, lines 86-213) and the cozystack mirror in `charts/cozystack/templates/_helpers.tpl` only emits a `VLANConfig` when:

1. `talm.discovered.default_link_name_by_gateway` resolves to a link, AND
2. that link has `spec.kind == "vlan"` per the live COSI `links` resource.

Three independent paths drop a configured VLAN:

1. **Secondary VLANs are dropped unconditionally.** The renderer handles only the gateway-bearing link. A VLAN that does not carry the default route (storage / mgmt) is never emitted, regardless of discovery state. See `charts/generic/templates/_helpers.tpl:121-184`. Tracked separately as a coverage gap.

2. **`existing_interfaces_configuration` is intentionally bypassed in multi-doc mode** (`charts/generic/templates/_helpers.tpl:87-90`). The legacy path consults `lookup "machineconfig" "" "v1alpha1"` and copies user-declared interfaces verbatim; the multi-doc path explicitly does not. Any VLAN declared in the existing machine config (legacy schema) is silently discarded. Tracked separately.

3. **Discovery may not see the VLAN at template time.** `is_vlan` reads `$link.spec.kind == "vlan"` from the live COSI `links` resource (`charts/talm/templates/_helpers.tpl:262-268`). If the VLAN link is not yet up (first apply, or replay after wipe, or the older Talos on the live node represents the link differently), the check fails and no `VLANConfig` is emitted.

## Information needed to narrow it down

- Output of `talm template ...` for the affected node (IPs may be redacted).
- `talosctl --nodes <NODE_IP> get links --output yaml` to confirm what discovery actually sees, and the value of `spec.kind` for the VLAN link.
- Talos version reported by `talosctl --nodes <NODE_IP> version` vs `talosVersion` declared in `values.yaml` / `Chart.yaml`.
- Where the VLAN was declared on the cluster: in the running `MachineConfig` (legacy `machine.network.interfaces[].vlans[]`), or already as a separate `VLANConfig` document?
- Whether the VLAN carries the default route.

### LINKED ISSUE #141: bug(chart): multi-doc renderer ignores user-supplied machine.network.interfaces
## What

On the Talos v1.12 multi-doc rendering path (`charts/generic/templates/_helpers.tpl:talos.config.network.multidoc`), the chart fully reconstructs network configuration from live discovery resources (`links`, `addresses`, `routes`) and explicitly does NOT consult `talm.discovered.existing_interfaces_configuration`.

Source comment (`charts/generic/templates/_helpers.tpl:87-90`):

> Multi-doc format always reconstructs network config from discovery resources.
> existing_interfaces_configuration is not used here because v1.12 nodes store
> network config in separate documents (LinkConfig, BondConfig, etc.), not in
> the legacy machine.network.interfaces field.

The premise — "v1.12 nodes already store network config in separate documents" — does not hold for clusters that were upgraded from a pre-1.12 chart. Their running `MachineConfig` still carries `machine.network.interfaces[]` from the legacy schema, and the multi-doc renderer drops it on first re-render.

## Symptom

A user upgrading from v0.23 to v0.24+ does not get an automatic migration of legacy `machine.network.interfaces[]` declarations into multi-doc form. The legacy declarations vanish from the rendered config.

This is invisible at template time: `talm template` produces a config that looks complete (HostnameConfig, ResolverConfig, one LinkConfig / VLANConfig / BondConfig for the gateway link) but is missing every secondary interface, every secondary VLAN, every static route, and every user-side override that lived under `machine.network.interfaces`.

## Reproduction shape (any of these triggers it)

- Node with two NICs, one routed (gateway link) and one for storage.
- Node with multiple VLANs, only one of which carries the default route.
- `values.yaml` that injects extra interfaces, addresses, or routes via the legacy `machine.network.interfaces[]` schema.

## Affected versions

Introduced in #116 (v0.24.0). Present in `main` as of writing.

## Possible directions

1. Consult `existing_interfaces_configuration` in multi-doc mode and translate non-trivial entries into `LinkConfig` / `VLANConfig` / `BondConfig` / `RouteConfig` documents.
2. Read the user-supplied source values (the `machine.network.interfaces[]` block in `values.yaml`) and translate them at render time.
3. As an immediate guardrail: hard-fail or emit a noisy warning when multi-doc mode would drop a non-empty `existing_interfaces_configuration`. Smallest change; unblocks users until a real translator is built.

Option 3 is recommended as the first step.

### LINKED ISSUE #142: bug(chart): multi-doc renderer emits network docs only for the gateway-bearing link
## What

On the Talos v1.12 multi-doc path, both `charts/generic/templates/_helpers.tpl:talos.config.network.multidoc` and the equivalent in `charts/cozystack/templates/_helpers.tpl` resolve a single link via `talm.discovered.default_link_name_by_gateway` and emit network documents (`LinkConfig` / `BondConfig` / `VLANConfig` / `Layer2VIPConfig`) only for that one link.

Every other physical NIC, bond, VLAN, or bridge present on the node is ignored by the renderer.

## Why this is a bug

Multi-NIC nodes are an expected shape; #125 tracks the discovery side ("discover secondary network links alongside primary gateway link"). The discovery helpers added in #127 (`physical_link_names`, `configurable_link_names`, `addresses_by_link`, `gateway_by_link`, `routes_by_link`) exist specifically to enumerate the full link inventory, but the multi-doc renderer does not consume them — it still resolves a single link and stops.

Result: on a node with one routed NIC plus a storage NIC, the rendered config configures the routed NIC and silently leaves the storage NIC unconfigured. Same for any non-default-route VLAN, additional bond, or bridge.

## Affected versions

Coverage gap since the multi-doc path was introduced in #116 (v0.24.0).

## Suggested direction

Iterate `configurable_link_names` and emit the appropriate document per link kind, populating addresses and routes via the by-link helpers from #127. The single-gateway-link logic becomes a special case (the link that carries the default route also gets the gateway route appended).

Related to #125, but distinct: that issue tracks discovery; this one tracks the chart-side rendering. Even with full discovery available today, the renderer picks one link.

### LINKED ISSUE #143: talm template --talos-version v1.12 does not discover VLAN interfaces from running node
## Description

When running `talm template` against a node that has a VLAN interface configured (e.g., `enp0s31f6.4000` via `machine.network.interfaces[].vlans`), the generated v1.12 multidoc config only includes a `LinkConfig` for the physical parent interface. The VLAN interface is silently omitted.

As a result:
- The VLAN private IP is missing from the generated config
- `Layer2VIPConfig` is generated with `link: enp0s31f6` (physical/public) instead of `link: enp0s31f6.4000` (VLAN) — placing the VIP on the wrong interface

## Steps to Reproduce

1. Bootstrap a node with a VLAN configured via legacy machine config:
   ```yaml
   machine:
     network:
       interfaces:
         - interface: enp0s31f6
           vlans:
             - vlanId: 4000
               addresses:
                 - 192.168.100.2/24
               vip:
                 ip: 192.168.100.10
   ```
2. Set `talosVersion: "v1.12"` in `Chart.yaml`
3. Run `talm template -t templates/controlplane.yaml -n <node-ip> -e <node-ip>`

## Observed Output

```yaml
---
apiVersion: v1alpha1
kind: LinkConfig
name: enp0s31f6
addresses:
  - address: 88.99.249.47/26
routes:
  - gateway: 88.99.249.1
---
apiVersion: v1alpha1
kind: Layer2VIPConfig
name: "192.168.100.10"
link: enp0s31f6       # wrong: should be enp0s31f6.4000
```

The VLAN interface (`enp0s31f6.4000`, `192.168.100.2/24`) is completely absent.

## Expected Behavior

The template should also generate VLAN configuration, either as a separate VLAN document (if such a kind exists in v1.12) or at minimum annotate the discovered interfaces comment to include VLANs, so the user knows they need to add it manually. `Layer2VIPConfig` should reference `enp0s31f6.4000` when the VIP is configured on a VLAN interface.

## Workaround

Manually add VLAN config via legacy format in the first document and fix the VIP link:

```yaml
machine:
  network:
    interfaces:
      - interface: enp0s31f6
        vlans:
          - vlanId: 4000
            addresses:
              - 192.168.100.2/24
---
apiVersion: v1alpha1
kind: Layer2VIPConfig
name: "192.168.100.10"
link: enp0s31f6.4000
```

## Environment

- talm: dev (post v0.25.0)
- Talos: v1.12.6
- VLAN ID: 4000, parent interface: enp0s31f6

lexfrei:
## What changed

The v1.12 multi-doc network renderer in cozystack and generic charts now emits one document per configurable link on the node instead of stopping at the gateway-bearing one. Multi-NIC nodes therefore produce multiple `LinkConfig` / `BondConfig` / `VLANConfig` documents — a node with a routed NIC plus a storage NIC ends up with both NICs configured in the rendered config.

The IPv4 default-route gateway is placed only on the link that carries it; every other link is emitted gateway-less. MTU is surfaced on `LinkConfig`, `BondConfig`, and `VLANConfig` when discovery reports a value, so non-default-MTU links (jumbo frames, GRE) survive a re-render. Both IPv4 and IPv6 global-scope addresses on a link are surfaced.

Guardrails:

- **Bond slaves** are filtered out of `configurable_link_names` so they do not collide with the master's `BondConfig`.
- **Bridges** are deliberately not auto-emitted as `BridgeConfig` yet — a non-gateway bridge is skipped (declare it via per-node body); a bridge that carries the default route fails the render with a clear migration hint.
- **VLAN with missing `parent` or `vlanID`** fails the render rather than emitting a document Talos rejects on apply.
- **Bond with partial `bondMaster`** gracefully degrades — unset fields are omitted instead of producing `bondMode: <nil>`.
- **`floatingIP` is stripped from per-link addresses** so the VIP currently held by a leader does not leak into the static `LinkConfig`.
- **Legacy `machine.network.interfaces[]` in the running MachineConfig** triggers a hard fail with a migration message, preventing silent loss of user network declarations during upgrade from a chart that emitted the legacy schema.

## Tests

12 new `MultiDoc` subtests pin every guardrail and every emit path: multi-link iteration, MTU on all three doc kinds, VLAN emission, bond slaves filtered, bridge skip + bridge-as-gateway fail, VLAN-without-parent fail, VLAN-without-vlanID fail, partial-bondMaster gracefully degraded, floatingIP stripped from addresses, dual-stack IPv6 addresses surfaced, Layer2VIPConfig.link on bond and VLAN gateway, legacy-interfaces guardrail. Six new fixtures mirror real Talos COSI shape (slaveKind, masterIndex, linkIndex). README updated to describe per-link emission, the bond-slave filter, the bridge contract, the dual-stack contract, the VIP-strip, and the upgrade-from-legacy guardrail.

Closes #140.
Closes #141.
Closes #142.
Closes #143.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,27 +8,15 @@ Talm is just like Helm, but for Talos Linux
 
 While developing Talm, we aimed to achieve the following goals:
 
-- **Automatic Discovery**: In a bare-metal environment, each server may vary
-slightly in aspects such as disks and network interfaces.
-Talm enables discovery of node information, which is then used to generate patches.
+- **Automatic Discovery**: In a bare-metal environment, each server may vary slightly in aspects such as disks and network interfaces. Talm enables discovery of node information, which is then used to generate patches.
 
-- **Ease of Customization**: You can customize templates to create your unique
-configuration based on your environment. The templates use the standard
-Go templates syntax, enhanced with widely-known Helm templating logic.
+- **Ease of Customization**: You can customize templates to create your unique configuration based on your environment. The templates use the standard Go templates syntax, enhanced with widely-known Helm templating logic.
 
-- **GitOps Friendly**: The patches generated do not contain sensitive data,
-allowing them to be stored in Git in an unencrypted, open format. For scenarios
-requiring complete configurations, the `--full` option allows the obtain
-a complete config that can be used for matchbox and other solutions.
+- **GitOps Friendly**: The patches generated do not contain sensitive data, allowing them to be stored in Git in an unencrypted, open format. For scenarios requiring complete configurations, the `--full` option allows the obtain a complete config that can be used for matchbox and other solutions.
 
-- **Simplicity of Use**: You no longer need to pass connection options for each
-specific server; they are saved along with the templating results into
-a separate file. This allows you to easily apply one or multiple files in batch
-using a syntax similar to `kubectl apply -f node1.yaml -f node2.yaml`.
+- **Simplicity of Use**: You no longer need to pass connection options for each specific server; they are saved along with the templating results into a separate file. This allows you to easily apply one or multiple files in batch using a syntax similar to `kubectl apply -f node1.yaml -f node2.yaml`.
 
-- **Compatibility with talosctl**: We strive to maintain compatibility with the upstream
-project in patches and configurations. The configurations you obtain can be used
-with the official tools like talosctl and Omni.
+- **Compatibility with talosctl**: We strive to maintain compatibility with the upstream project in patches and configurations. The configurations you obtain can be used with the official tools like talosctl and Omni.
 
 
 ## Installation
@@ -52,14 +40,7 @@ curl -sSL https://github.com/cozystack/talm/raw/refs/heads/main/hack/install.sh 
 
 ### Windows
 
-Windows is supported. Download the `talm-windows-*.zip` archive from the
-[releases page](https://github.com/cozystack/talm/releases/latest) and
-extract `talm.exe`. On Windows, template paths passed to the `-t` /
-`--template` flag accept either `\` or `/` separators, so
-`-t templates\controlplane.yaml` and `-t templates/controlplane.yaml`
-are equivalent. Other path flags (`--talosconfig`, `-f` / `--file`)
-are delegated to the underlying OS file loader and follow standard
-Windows path rules.
+Windows is supported. Download the `talm-windows-*.zip` archive from the [releases page](https://github.com/cozystack/talm/releases/latest) and extract `talm.exe`. On Windows, template paths passed to the `-t` / `--template` flag accept either `\` or `/` separators, so `-t templates\controlplane.yaml` and `-t templates/controlplane.yaml` are equivalent. Other path flags (`--talosconfig`, `-f` / `--file`) are delegated to the underlying OS file loader and follow standard Windows path rules.
 
 ## Getting Started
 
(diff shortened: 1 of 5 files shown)


kitsunoff (approved):
LGTM

PR #148, opened 2026-05-07
feat(charts): allow overriding of the cluster's name

dislogical:
This allows users to override the generated cluster name in the case they don't want it to match the chart name.

## Summary by CodeRabbit

* **New Features**
  * Customizable cluster naming in Helm charts with automatic fallback to the chart name and enforced DNS-1123 validation (invalid names now produce an error).

* **Configuration**
  * Added optional clusterName chart value to override the default name.

* **Tests**
  * Rendering tests updated to validate override behavior and enforce quoted clusterName output.

### DIFF
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -103,7 +103,7 @@ cluster:
       {{- toYaml .Values.podSubnets | nindent 6 }}
     serviceSubnets:
       {{- toYaml .Values.serviceSubnets | nindent 6 }}
-  clusterName: "{{ .Chart.Name }}"
+  clusterName: {{ .Values.clusterName | default .Chart.Name | regexFind "^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$" | required "clusterName must be a valid DNS-1123 label" | quote }}
   controlPlane:
     endpoint: {{ required "values.yaml: `endpoint` must be set to the cluster control-plane URL (e.g. https://<vip>:6443). This field is cluster-wide: every node's kubelet and kube-proxy dials it, so it cannot be auto-derived from the current node's IP -- `talm template` runs once per node and has no way to reconcile per-node IPs into a single shared endpoint. For multi-node setups use a VIP (cozystack floatingIP) or an external load balancer; for single-node clusters the node's routable IP works." .Values.endpoint | quote }}
   {{- if eq .MachineType "controlplane" }}
--- a/charts/cozystack/values.yaml
+++ b/charts/cozystack/values.yaml
@@ -19,6 +19,12 @@
 # Example: endpoint: "https://192.168.0.1:6443"
 endpoint: ""
 
+# Optional override for the cluster's name (defaults to Chart.Name).
+# Note that changing this value on a live cluster is considered
+# dangerous as it is baked into PKI (cert SANs) and ETCD identity at
+# bootstrap.
+clusterName: ""
+
 clusterDomain: cozy.local
 # Layer-2 VIP for cozystack multi-node setups. When set, the chart
 # emits a Layer2VIPConfig document pinning this IP as a floating
--- a/charts/generic/templates/_helpers.tpl
+++ b/charts/generic/templates/_helpers.tpl
@@ -45,13 +45,14 @@ machine:
 
 {{- /* Shared cluster section */ -}}
 {{- define "talos.config.cluster" }}
+
 cluster:
   network:
     podSubnets:
       {{- toYaml .Values.podSubnets | nindent 6 }}
     serviceSubnets:
       {{- toYaml .Values.serviceSubnets | nindent 6 }}
-  clusterName: "{{ .Chart.Name }}"
+  clusterName: {{ .Values.clusterName | default .Chart.Name | regexFind "^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$" | required "clusterName must be a valid DNS-1123 label" | quote }}
   controlPlane:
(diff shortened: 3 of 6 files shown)


lexfrei (changes requested):
NOT LGTM — the override is wired only into the chart templates; the `talm talosconfig` regeneration path still reads the cluster name from `Chart.yaml`, and the new behaviour ships without test coverage that pins either the legacy default or the override precedence.

**Business context**: lets users override the cluster name without renaming the chart directory or re-running `talm init --name`. The previous escape hatch (`--name` baked into `Chart.yaml` at init) requires re-init; this PR makes the override mutable post-init via `values.yaml`.

## Blockers

### B1: `talm talosconfig` regenerates with the chart name, not the new override

**File**: `pkg/commands/talosconfig.go:131-137` (interaction with `charts/{cozystack,generic}/templates/_helpers.tpl`)

When `regenerateTalosconfig` runs without an existing `talosconfig` (fresh checkout, deleted file, or first `talm talosconfig` after bootstrap loss), it falls back to `getClusterNameFromChart()`, which reads only `Chart.yaml.name`. The rendered Talos machine config now has `cluster.clusterName: <values.clusterName>`, but the regenerated talosconfig context is `<Chart.Name>`.

`getClusterNameFromChart()` at `pkg/commands/talosconfig.go:194-210` parses only `Chart.yaml`; it never opens `values.yaml`. The path is reachable via the user-facing `talm talosconfig` command (`talosconfig.go:34-82`), which is the documented "regenerate after cert expiry" flow.

**Impact**: silent cluster-name divergence between client and server config whenever the override is in use and the regenerate path is hit.

**Fix**: either (a) extend `getClusterNameFromChart()` to also load `values.yaml` and prefer `clusterName` when non-empty, or (b) call out in the PR description and the `values.yaml` comment that the override is template-only and the user must keep `Chart.yaml.name` in sync.

### B2: Missing test coverage for the new behaviour

**File**: `pkg/engine/render_test.go`

Current tests (lines 231, 292, 345, 507) only assert the `clusterName:` substring is present — none of them pin the rendered value. Both the legacy fallback and the new override are uncovered, and the change touches two charts (`cozystack` and `generic`) with independent template files, so a "fixed cozystack, forgot generic" regression would slip through CI. Minimum coverage expected before merge:

**Backward compatibility (legacy contract — must not regress)**

1. `clusterName: ""` (explicit empty in values) → rendered output contains `clusterName: "<Chart.Name>"`. Confirms the `default` pipeline restores the prior literal.
2. `clusterName` key entirely absent from values → same `clusterName: "<Chart.Name>"`. Guards against a future `values.yaml` cleanup dropping the key and silently breaking the fallback.

**Override precedence**

3. `clusterName: "my-cluster"` → rendered output contains `clusterName: "my-cluster"`, NOT `<Chart.Name>`. Confirms the override beats the default.

**Edge cases that pin current behaviour**

4. Special characters / spaces: `clusterName: "foo bar"` → output contains `clusterName: "foo bar"` (quoted). Confirms `quote` actually fires.
5. Override equal to `Chart.Name`: idempotent — same output as case 1.
6. YAML-ambiguous string values: `clusterName: "true"` and `clusterName: "123"` → output keeps them as quoted strings, not booleans/ints. Without `quote` these would round-trip as native YAML scalars.
7. Whitespace-only: `clusterName: "   "` → output is `clusterName: "   "` (sprig `default` does NOT treat whitespace as empty). Either pin this behaviour with a comment explaining why, or add a `trim` in the template — but the test must exist either way.

**Both charts**

Cases 1–3 must be exercised against BOTH `charts/cozystack` and `charts/generic` (the diff touches both `_helpers.tpl` files independently). A single shared helper executed twice with different chart roots is fine.

The existing `pkg/engine/render_test.go` fixtures already provide an inline-values pattern; the simplest delivery is to convert one of the existing four `assertContains(t, output, "clusterName:")` call sites into a table-driven subtest covering the cases above.

## Non-blocking follow-ups

1. The `values.yaml` comment ("Optional override for the cluster's name (defaults to Chart.Name)") understates the implication: `cluster.clusterName` is baked into PKI (cert SANs) and ETCD identity at bootstrap. Changing it on a live cluster is unsafe. A one-sentence warning in the comment would prevent the typical "I'll just rename my cluster" mistake.
2. No DNS-1123 validation. A non-conforming string (uppercase, spaces, special chars) propagates through and fails downstream in Talos with a less obvious error. An optional Helm `fail` guard could surface this at template time.
3. README's only mention of `clusterName` is in a sample output snippet (line 154). Neither the existing `--name` init flag nor the new override are documented as user-facing knobs. Pre-existing gap, not made worse by this PR.

dislogical:
Thanks for feedback! I've addressed all comments except non-blocking 3, which seems out of scope for this work.

lexfrei (approved):
LGTM — both prior blockers are addressed: `getClusterNameFromChart` now reads `values.yaml.clusterName` ahead of `Chart.yaml.name`, and the test suite pins the override behaviour with both positive and negative cases on each chart.

**Business context**: lets users override the cluster name without renaming the chart directory or re-running `talm init --name`.

## What changed since the prior review

- `pkg/commands/talosconfig.go:194-219` — `getClusterNameFromChart` now resolves `values.yaml.clusterName` first, falls back to `Chart.yaml.name`. Closes the regenerate-path divergence flagged in the previous review.
- `pkg/engine/render_test.go` — four new tests (`TestMultiDoc{Cozystack,Generic}_{Valid,Invalid}ClusterNameOverride`) and four existing tests upgraded from substring-only `clusterName:` to value-pinning `clusterName: "<expected>"`. Closes the test-coverage blocker on the chart side.
- `charts/{cozystack,generic}/values.yaml` — comment now warns that changing the value on a live cluster is dangerous (PKI / ETCD identity baked at bootstrap).
- `charts/{cozystack,generic}/templates/_helpers.tpl` — DNS-1123 regex via `regexFind ... | required` rejects invalid names at render time.

## Non-blocking follow-ups

1. No unit test in `pkg/commands` covers the new `values.yaml.clusterName > Chart.yaml.name` priority in `getClusterNameFromChart`. The chart-side tests in `render_test.go` exercise the template, not the regenerate-talosconfig flow. A short test that creates a tmp dir with `values.yaml: clusterName: foo` + `Chart.yaml: name: bar` and asserts the helper returns `foo` would close the gap.
2. The DNS-1123 regex does not enforce the 63-character label length cap (RFC 1035). A name longer than 63 chars passes the chart validation but Talos / Kubernetes downstream may still reject it. Minor; the regex captures the practical typo cases (uppercase, underscore, leading dash, etc.) which is the high-value fraction.

dislogical:
Thank you!

PR #196, opened 2026-05-12
[codex] Improve dmesg tail flag error

### LINKED ISSUE #195: talm dmesg: --tail=N produces a cryptic ParseBool error; cushion the footgun in the wrapper
`talm dmesg --tail=3` fails with:

```text
invalid argument "3" for "--tail" flag: strconv.ParseBool: parsing "3": invalid syntax
```

This is upstream behaviour: `talosctl` defines `--tail` as a boolean toggling tail-mode for the `--follow` stream, not as a line count. The gRPC method signature `Dmesg(ctx, follow, tail bool)` reflects the same shape; there's no server-side line-count parameter to forward to.

Operator confusion source: the flag name `--tail` mirrors `tail(1)`'s `-n <N>` semantics, which is the default mental model when an operator wants "the last N lines". An operator typing `--tail=3` is asking for line count, not tail-mode. The `strconv.ParseBool` error message gives no clue about what `--tail` actually does.

Upstream classifies as deliberate API — not a bug, just consistent gRPC shape. The flag works exactly as documented in its help text. But the footgun lands operators in a cryptic error and is exactly the class of thing talm should help defuse.

## Options

1. **Pre-flight hint in the talm wrapper.** Detect `--tail=<value-that-fails-ParseBool>` and rewrite the error with a clear hint:

    ```text
    talm dmesg: --tail is a boolean toggling tail-mode for --follow, not a line count.
    For the last N lines, run:
        talm dmesg --nodes <node> | tail -n 3
    To stream only new messages on a follow:
        talm dmesg --follow --tail
    ```

2. **Client-side `--tail-lines=N` flag.** A new talm-owned flag that buffers the dmesg stream and emits only the last N lines client-side. Cleaner UX, but introduces talm-specific surface distinct from upstream; future readers of `talm dmesg --help` see flags that don't exist in `talosctl dmesg`.

3. **Documentation-only.** Update `--help` for the wrapped `dmesg` command to explicitly note that `--tail` takes no value and that line-count lives in pipe-to-`tail`. Cheapest, doesn't help operators who don't read help before typing.

Option 1 is the lowest-cost and addresses the actual harm: the operator gets a clear next step instead of a cryptic ParseBool error. Option 2 gives the best UX but accepts talm-specific divergence from upstream.

## Why not upstream

The upstream definition is internally consistent: `BoolVarP(&dmesgTail, "tail", "", false, "specify if only new messages should be sent (makes sense only when combined with --follow)")` in [siderolabs/talos — cmd/talosctl/cmd/talos/dmesg.go](https://github.com/siderolabs/talos/blob/main/cmd/talosctl/cmd/talos/dmesg.go). The gRPC `Dmesg(ctx, follow, tail bool)` API matches. Filing against siderolabs/talos would ask them to either rename a stable flag (breaking) or break their wire API (also breaking). Mitigation belongs in talm's wrapper layer.

## Reproducer

```text
$ talm dmesg --tail=3
Error: invalid argument "3" for "--tail" flag: strconv.ParseBool: parsing "3": invalid syntax
$ echo $?
1
```

stellhub:
Fixes #195.

## Summary

- Replaces the `strconv.ParseBool`-style `talm dmesg --tail=N` failure with a talm-specific actionable error.
- Adds hints for piping to `tail -n N` and for the boolean `--follow --tail` streaming use case.
- Keeps unrelated Cobra flag errors unchanged.

## Validation

- `$env:GOSUMDB='sum.golang.org'; go test ./pkg/commands`
- `$env:GOSUMDB='sum.golang.org'; go build -o .\talm-test.exe .`
- `.\talm-test.exe dmesg --tail=3` prints:
  - `talm dmesg: --tail is a boolean toggling tail-mode for --follow, not a line count`
  - `hint: for the last 3 lines, run: talm dmesg --nodes <node> | tail -n 3`
  - `hint: to stream only new messages on a follow, run: talm dmesg --follow --tail`

Also tried `$env:GOSUMDB='sum.golang.org'; go test ./...` locally on Windows. Packages outside `pkg/engine` passed, but `pkg/engine` failed before reaching this change because the checked-out Git symlink files `charts/cozystack/charts/talm` and `charts/generic/charts/talm` are plain 10-byte files on this Windows checkout, so Helm reports `Chart.yaml file is missing` for those chart tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for the `dmesg` command when `--tail` is used with invalid values, providing clearer guidance to users.

* **Tests**
  * Added contract tests to validate improved error messages for `dmesg` command argument parsing and ensure other flag errors remain unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/talm/pull/196)

### DIFF
--- a/pkg/commands/talosctl_wrapper.go
+++ b/pkg/commands/talosctl_wrapper.go
@@ -172,6 +172,11 @@ func wrapTalosCommand(cmd *cobra.Command, cmdName string) *cobra.Command {
 		wrapUpgradeCommand(wrappedCmd, originalRunE)
 	}
 
+	// Special handling for dmesg command
+	if baseCmdName == "dmesg" {
+		wrapDmesgCommand(wrappedCmd)
+	}
+
 	// Special handling for rotate-ca command
 	if baseCmdName == "rotate-ca" {
 		wrapRotateCACommand(wrappedCmd, originalRunE)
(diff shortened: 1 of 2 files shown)


lexfrei:
Hi @stellhub — apologies for the timing collision. I iterated on this very quickly and didn't surface the duplicate before you put work into it: a `wrapDmesgCommand` cushion already landed in main via #197 (commit c2f3017, May 12) and that's what closed #195 a few hours before your PR went up. I genuinely appreciate the time you spent on a clean implementation with extracted helpers and contract-style tests — the shape is good.

On top of the duplication, the existing cushion (and consequently the hint in your PR) is going to be rewritten anyway: the upstream maintainer's response on siderolabs/talos#13333 points at `talosctl logs kernel --tail=N` as the proper "last N kernel lines" path (and notes that `dmesg` itself is on track for retirement). The right hint is now "use `talm logs kernel --tail=N`", not the `... | tail -n N` pipe workaround that both versions currently offer. I'm opening a follow-up PR to rewrite the existing cushion against that guidance.

Closing this PR. Thanks again for the effort and the careful test coverage — please don't let this timing turn you off contributing more.

PR #210, opened 2026-05-15
feat(charts,modeline,upgrade): operator extension points and dev17 install fixes

lexfrei:
## Summary

Three independent improvements surfaced while installing Cozystack on a fresh 3-node Talos cluster (`dev17`, OCI VM.Standard.E5.Flex, Talos v1.12.6 → v1.13.0). Each commit stands alone and can be reverted independently.

### `fix(modeline): split modeline parts on top-level commas only`

The `# talm: nodes=[...], endpoints=[...]` parser used `strings.SplitSeq(content, ", ")` which cut the value at the first comma INSIDE a multi-element JSON array. Hand-authored shared side-patches with `nodes=["a", "b"]` (space after comma) hit `error parsing JSON array for key nodes, value ["a"`. The talm-generated form happened to dodge it because `json.Marshal` of a string slice emits no space after the comma.

Replace the literal split with a depth-aware tokenizer that splits on `,` only at JSON-array depth 0, tracking string state and backslash escapes. Both the canonical no-space form and the human form parse cleanly. Comma inside string literals never splits. Empty arrays, trailing commas, and unbalanced brackets keep their existing semantics (test-pinned).

### `feat(charts): expose Talos operator extension points in cozystack + generic presets`

Operators wanting to extend the rendered Talos config (`machine.kernel.modules`, `machine.sysctls`, `machine.kubelet.extraConfig`, `machine.files`) had to fork the chart or hand-edit the generated `nodes/<n>.yaml` against the autogen banner. Adds four `extra*` values keys to both presets.

On cozystack (defaults present): list-shaped keys append; map-shaped keys are collision-checked at template time and `fail` with a precise hint if an operator key names a built-in. Built-ins are never overridable — yaml.v3 rejects duplicate map keys on decode, so a silent emit-both would produce a config that cannot round-trip.

On generic (no defaults): each block emits only when the matching `extra*` key is non-empty.

### `fix(upgrade): point-patch node body install.image after successful upgrade`

`talm upgrade` resolves the target installer image from `values.yaml`, but the per-node `nodes/<n>.yaml::machine.install.image` was never resynced afterwards. A follow-up `talm apply` rendered the chart against `values.yaml` (got the new image), then merged the stale body via `MergeFileAsPatch` and silently pinned the cluster back to the pre-upgrade image — the next A/B boot would roll back without warning.

After successful RPC + post-upgrade verify, point-patch `machine.install.image` in every `-f` node body to the applied image. Uses a yaml.v3 streaming decoder so the v1.12+ multi-doc shape (machine + cluster + RegistryMirrorConfig + LinkConfig + Layer2VIPConfig + VLANConfig) is preserved intact — `yaml.Unmarshal` reads only the first document and would silently erase the rest. Encoder pinned at 2-space indent so the diff stays a single line. Comments and modeline survive untouched.

Verify-failure (auto-rollback) intentionally leaves the body alone: the node ran the old image, so the body must reflect that. Operator opt-out via `--skip-post-upgrade-verify` (and `--insecure` / `--stage` paths that drop Phase 2C entirely) patches unconditionally on RPC success.

## Verification

- `go test ./...`: all packages green.
- `golangci-lint run ./...`: 0 issues.
- Manual-test-plan extended in each commit (B7/B8 for extension points, C6f for multi-IP modeline, E4 for post-upgrade body sync).

## Docs sync

The new `extra*` keys and the post-upgrade body-sync behaviour are operator-visible. A parallel update at `cozystack/website` will follow under a separate PR.


lexfrei:
Closing in favour of a follow-up PR with a cleaner branch name and description.

PR #214, opened 2026-06-02
feat(version): warn when project charts drift from the talm binary version

myasnikovdaniil:
## What

talm vendors its preset and library charts into the project directory at
`talm init` time. Render commands (`template`/`apply`/`upgrade`) load that
**local** copy — never the binary's embedded charts. So upgrading the talm
binary leaves `charts/talm/` frozen at whatever version last ran init, and
that drift was invisible: operators had to manually eyeball `talm version`
against `Chart.yaml`.

This PR surfaces the drift.

## How

- Compare the binary version against the `version:` stamped in the project's
  `Chart.yaml` on every config-loading command.
- **Default:** non-fatal `WARN:` on stderr pointing at `talm init --update`.
  Rendered output (stdout) and exit code are unchanged — pipelines/CI keep working.
- **Opt-in enforcement:** `strictVersion: true` in `Chart.yaml` (committed, so a
  whole team/CI inherits it) or the `--strict-version` flag turns a mismatch
  into a hard error (exit 1).
- Stays **silent** for `dev` builds, the `0.1.0` dev sentinel, empty/unparseable
  versions, and matching versions — never a false alarm.

Reuses the existing `version:` stamp written by `init`/`init --update`; no new
required field. `strictVersion` is optional and absent means today's behavior —
fully backward-compatible.

## Docs

- New README section *"How talm vendors charts (keeping a project in sync)"*
  explaining the embed→vendor→render model and the `init --update` workflow.
- Extended manual-test-plan B5 with the warning + strict-mode assertions.

## Testing

- `go test ./...` green; `go vet ./...` clean.
- New unit tests (`version_check_test.go`, 9 cases) cover every warn/skip path.
- Verified end-to-end with real `-ldflags` builds: warn / strict-via-Chart.yaml
  / strict-via-flag → exit 1, dev-build → silent, matched → silent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added chart version drift detection that warns when the talm binary and vendored chart versions diverge
  * Added `--strict-version` flag and configuration option to enforce strict version matching with hard errors

* **Documentation**
  * Documented chart vendoring behavior and procedures for keeping projects synchronized after talm binary updates
  * Added testing documentation for drift warning and strict mode functionality


lexfrei:
Thanks for digging into this — the underlying problem is real: `talm` renders from the project's vendored `charts/talm/`, so upgrading the binary leaves that copy frozen and the drift is invisible.

I'm closing this in favor of a different implementation, because comparing the binary version against the project's top-level `Chart.yaml` `version:` has two issues:

1. **Wrong file.** What actually drifts is the vendored library (`charts/talm/`), not the project's own chart version — which an operator may bump on their own cadence, making the warning fire on every command indefinitely.
2. **Version number, not content.** The library `version:` is stamped with the binary version at init regardless of whether any chart file changed, so a patch bump that left `charts/talm/` byte-identical would still warn — a false positive.

The replacement detects drift by comparing the vendored `charts/talm/` against the binary's embedded copy **by content** (normalizing the version stamp away), so a pure version bump is never flagged. It keeps the UX you proposed: a non-fatal stderr warning by default, opt-in strict mode via `Chart.yaml` and a `--strict-charts` flag. Replacement: #216.

Appreciate the contribution — it surfaced a genuine gap.

PR #215, opened 2026-06-03
fix(cozystack): exclude loop devices from LVM global_filter

kvaps:
Add `/dev/loop` to the LVM `global_filter` in the generated Talos machine config so the host does not scan or activate volume groups located inside loop-mounted images.

## Summary by CodeRabbit

* **Chores**
  * LVM configuration updated to filter out loop devices (now excludes /dev/loop.* alongside existing patterns).
* **Tests**
  * Expanded test coverage to assert the LVM configuration includes the loop-device exclusion.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/talm/pull/215?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -110,7 +110,7 @@ machine:
         archive = 0
       }
       devices {
-         global_filter = [ "r|^/dev/drbd.*|", "r|^/dev/dm-.*|", "r|^/dev/zd.*|" ]
+         global_filter = [ "r|^/dev/drbd.*|", "r|^/dev/dm-.*|", "r|^/dev/zd.*|", "r|^/dev/loop.*|" ]
       }
   {{- with .Values.extraMachineFiles }}
   {{- toYaml . | nindent 2 }}
--- a/pkg/engine/contract_machine_test.go
+++ b/pkg/engine/contract_machine_test.go
@@ -630,8 +630,10 @@ func TestContract_Machine_CertSANs_LoopbackUnconditional_Cozystack(t *testing.T)
 //     plugin paths. Required for SR-IOV / GPU device plugins to
 //     surface inside privileged containers.
 //  2. /etc/lvm/lvm.conf — disables LVM backup/archive and sets a
-//     global_filter that excludes drbd, dm-, zd- devices. Required so
-//     LVM does not race the storage stack at boot.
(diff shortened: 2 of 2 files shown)


lexfrei (changes requested):
NOT LGTM — the config change itself is correct and safe, but it adds a fourth entry to a list that is pinned by a contract test without extending that test, and it leaves the test's own description of the filter stale.

**Business context**: prevent the Talos host LVM from scanning/activating volume groups that live inside loop-mounted images (e.g. guest disk images), by rejecting `/dev/loop.*` in `global_filter` — consistent with the existing drbd/dm/zd rejects and with how the disk selector already treats loop devices as a virtual class.

## Blockers

### B1: the new `global_filter` loop entry is not pinned by its contract test, and the test comment is now stale

**File**: `pkg/engine/contract_machine_test.go:651-653` (assertions) and `:632-634` (comment)

**Issue**: `TestContract_Machine_Files_Cozystack` pins each filter entry individually — `r|^/dev/drbd.*|`, `r|^/dev/dm-.*|`, `r|^/dev/zd.*|` — but this PR adds `r|^/dev/loop.*|` to the template (`charts/cozystack/templates/_helpers.tpl:113`) without adding the matching assertion. The test's doc comment still reads "global_filter that excludes drbd, dm-, zd- devices", which no longer matches what the template emits.

**Evidence**: `go test ./pkg/engine/ -run TestContract_Machine_Files_Cozystack` passes on this branch. Because the test asserts only the three original entries, removing `r|^/dev/loop.*|` from the template would not fail any test — the new behavior is unpinned, while all three sibling entries are pinned.

**Impact**: a later edit that drops the loop reject would silently regress (the host LVM resumes scanning loop devices, reintroducing the bug this PR fixes) with a green test suite; the stale comment also misdescribes the shipped contract.

**Fix**: in the same commit, add an `assertContains` for `r|^/dev/loop.*|` right after the `zd` assertion, and update the comment to read "excludes drbd, dm-, zd-, loop devices".

lexfrei (approved):
LGTM — the contract-test gap from the previous review is resolved.

`TestContract_Machine_Files_Cozystack` now pins `r|^/dev/loop.*|` alongside drbd/dm/zd, and the contract comment was refreshed to match. Verified that the new assertion fails if the loop reject is dropped from the template, so the filter is now regression-protected. Pushed directly to the branch as a separate commit.

PR #217, opened 2026-06-05
fix(engine): require a template before rendering the chart

lexfrei:
## What

`talm template` with no `--file`/`--template` rendered the entire chart before checking that any template was actually requested. The render was wasted (its output is discarded), and online it fired live discovery first — so an operator who simply forgot `--file` saw a misleading `connection refused` from a node dial instead of the actionable `templates are not set` error.

## How

Move the empty-`TemplateFiles` guard ahead of the chart load and render, after the online multi-node precondition (which does no network I/O). The same error is now raised before any render or node dial, in both offline and online modes.

## Testing

- New contract test pins that the guard fires before rendering: a chart body that errors if executed stays unrendered.
- Full engine and module suite green.

### DIFF
--- a/pkg/engine/contract_render_test.go
+++ b/pkg/engine/contract_render_test.go
@@ -56,6 +56,33 @@ func TestContract_Render_NoTemplateFilesError(t *testing.T) {
 	}
 }
 
+// Contract: the empty-TemplateFiles guard fires BEFORE any chart
+// rendering. When no template is requested, talm must surface the
+// actionable "templates are not set" error without first rendering the
+// chart — otherwise a render-time failure (here a `fail` directive,
+// online a live `lookup` against an unreachable node) masks the real
+// problem with a misleading error. The chart body below errors if it is
+// ever executed; the guard must short-circuit before that happens.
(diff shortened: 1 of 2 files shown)


myasnikovdaniil (approved):
LGTM — the empty-templates guard is correctly moved ahead of chart load and render while staying after the network-free multi-node precondition, so `talm template` without `--file`/`--template` now fails fast with "templates are not set" instead of a misleading `connection refused` from a wasted discovery dial.

Verified: `FailIfMultiNodes` only reads `metadata.FromOutgoingContext` (no network I/O); the node dial lives in the `lookup` closure invoked only by `eng.Render`, which is now reached after the guard, so offline and online both raise the same error before any render or dial. The error string is byte-identical, just relocated. The only production caller that reaches the empty-templates path is `talm template` — `apply.go` renders only when `len(modelineTemplates) > 0` — so no regression. The contract test genuinely pins the ordering (a `{{ fail }}` chart body stays unrendered). `go test ./pkg/engine/...` and `go build ./...` pass.

PR #222, opened 2026-06-18
feat: first-class encrypted user values, honored at template and apply

### LINKED ISSUE #221: feat: first-class encrypted user values — kept out of committed node files, honored at both template and apply
## Summary

talm already encrypts its *fixed* secret set (`secrets.yaml`, `talosconfig`,
`kubeconfig`) with the age key in `talm.key`. There is, however, no first-class
way to handle **arbitrary user secret values** that templates consume. As a
result such secrets either land in git in plaintext, or cannot be supplied at
`apply` time at all.

## Use case

We inject secrets into the machine config that must be (a) stored **encrypted
at rest** in git and (b) **never** present in plaintext in committed rendered
node files. Concrete examples:

- a KMS plugin's Vault **AppRole `secret-id`** (passed as env to a static pod in
  `machine.pods`);
- a **private registry password** (`RegistryAuthConfig`).

These are real secret *values* referenced by chart templates — not Talos' own
PKI/tokens that `talm.key` already covers.

## Current behavior / gaps

1. Put the secret in `values.yaml` → committed in **plaintext**.
2. Put it in a separate values file referenced by templates → `talm template`
   honors `templateOptions.valueFiles`, but `talm template -I` then **bakes the
   secret into the committed `nodes/*.yaml` in plaintext**.
3. `talm apply` re-renders from the template (per the modeline, see #126) using
   **`values.yaml` only** — it does **not** honor `templateOptions.valueFiles`
   and has no `--values` / `--set-file`. So a value supplied only via a value
   file is **lost at apply** (renders empty / `required` fails), even though it
   rendered fine under `talm template`. This template↔apply inconsistency is the
   core blocker.

## Current workaround (and why it's awkward)

We keep all secret-bearing config out of the chart and pass it as a SOPS-
encrypted **side-patch**: `talm apply -f nodes/cpN.yaml -f secret-patch.yaml`.
Downsides:

- list fields (`machine.pods`, `cluster.apiServer.extraVolumes`) **concatenate**
  across layers, so the resource must live in exactly one layer — we can't
  template the static pod and inject just the creds via a patch (it duplicates);
- the side-patch ends up duplicating otherwise-static config;
- it's a manual, easy-to-forget extra `-f`.

## Proposal (any one would solve it)

1. **Honor `valueFiles` at `apply`** (and add `--values` / `--set-file`) so
   values are consistent between `talm template` and `talm apply`.
2. **Native encrypted user values:** extend the existing age workflow to an
   arbitrary user values file, e.g. `values-secret.yaml` ↔
   `values-secret.encrypted.yaml` (committed), decrypted via `talm.key` for both
   `template` and `apply`. This mirrors how `secrets.yaml` already works.
3. **Keep secret values out of committed rendered node files** — e.g. a way to
   mark values as secret so `talm template -I` writes a placeholder while the
   real value is injected only at apply.

## Environment

- talm v0.31.0, Talos v1.13, multi-doc render path.
- Related: #126 (apply re-renders from template per modeline), #189 (redact
  secrets in drift output).

lexfrei:
## Summary

Adds first-class support for arbitrary encrypted user secret values that chart templates consume (a registry password, a KMS plugin's secret-id, etc.) — encrypted at rest in git with the existing `talm.key`, decrypted in memory at render time, and kept out of committed node files and terminal output.

Closes #221.

## What changed

**`talm apply` honors all value sources.** apply re-renders from the modeline at apply time but previously had no `--values`/`--set*` flags and ignored `templateOptions.{valueFiles,values,...}` from Chart.yaml — so a value supplied at template time was silently dropped at apply. apply now registers the full value-source flag set and merges Chart.yaml's templateOptions, rendering identically to `talm template`. Chart.yaml-declared value files resolve against the project root, so they work from any CWD.

**Encrypted user value files, decrypted in memory.** A value file named `*.encrypted.yaml` (referenced from `templateOptions.valueFiles`) is age-decrypted in memory via `talm.key` at both template and apply — plaintext never touches disk. `talm init --encrypt`/`--decrypt` now covers `values-secret.yaml` ↔ `values-secret.encrypted.yaml`, and the plaintext is git-ignored. A misnamed/envelope-less `*.encrypted.yaml` fails loudly; a missing key surfaces a recovery hint.

**Secrets kept out of committed and printed output.** `talm template -I` omits secret-bearing fields from `nodes/*.yaml` (the real value is re-injected only at apply, which re-renders from the encrypted file). `talm template` stdout and `talm apply`'s drift preview redact them by default; `--show-secrets` / `--show-secrets-in-drift` reveal them. Omission is structural (whole map key or whole list element — never a partial list, which would clobber a `merge:"replace"` list at apply) and deterministic (no git churn).

## Testing

`go build`, `go test ./...`, `golangci-lint run` all clean. New contract tests cover the omit→merge round-trip (including `merge:"replace"` and empty-map cases), in-memory decryption, template↔apply value/path-resolution parity, drift redaction by value, the unpersisted-file warning, and the init encrypt/decrypt round-trip. `docs/manual-test-plan.md` and README updated.

## Summary by CodeRabbit

* **New Features**
  * Added encrypted user values via `values-secret.encrypted.yaml`, supporting template and apply flows with in-memory decryption.
  * `talm apply --dry-run` drift preview and post-apply verification now redact user-encrypted secrets by default, with `--show-secrets-in-drift` to reveal them.
  * Encrypted Helm `--values` files (`*.encrypted.yaml`) are decrypted in-memory during apply/template rendering.
  * Added `talm init` encryption/decryption support for `values-secret.yaml` plus gitignore updates.

* **Documentation**
  * Updated README and manual test plan with the encrypted user values workflow and clarified redaction/dry-run behavior.


sircthulhu:
### Drift preview doesn't redact encrypted-user-file values by value

Heads-up on a small gap I hit while using this branch (built on top of a local v1.14-machinery bump, but the redaction code path here is machinery-independent).

A secret value sourced from an encrypted user value file (`*.encrypted.yaml` via `templateOptions.valueFiles`) is **redacted in `talm template` stdout** (`***`), but the **same value appears verbatim in the `talm apply --dry-run` drift preview** — with no `--show-secrets-in-drift` passed. The PR text says the drift redacts these "by value … at any path", so it looks like the by-value redaction isn't applied on the drift path (only the static path-allowlist redaction seems to run there).

#### Minimal repro

`values-secret.yaml` → `talm init --encrypt` → `values-secret.encrypted.yaml`, referenced from `Chart.yaml`:

```yaml
# values-secret.yaml
demoSecret: "high-entropy-value-do-not-collide-abcdef0123456789"
```

A template rendering it into a non-allowlisted field (e.g. a static pod env):

```yaml
machine:
  pods:
    - apiVersion: v1
      kind: Pod
      metadata: {name: demo, namespace: kube-system}
      spec:
        containers:
          - name: demo
            image: busybox
            env:
              - name: DEMO
                value: {{ .Values.demoSecret | quote }}
```

Then:

- `talm template -f nodes/<node>.yaml` → `value: '***'` ✅ redacted
- `talm apply -f nodes/<node>.yaml --dry-run` → drift shows `value: high-entropy-value-…` ❌ verbatim

Expected: the drift preview redacts it the same way `talm template` does, unless `--show-secrets-in-drift` is set.

lexfrei:
Confirmed and fixed — thanks for the precise repro. You were right, and it's a surface my redaction didn't cover.

`talm apply --dry-run` emits two diffs. The first is talm's own structured drift preview, which redacts the value as expected. The second is the server-returned `Config diff:` block (Talos's `ModeDetails`), which talm was printing verbatim — so the secret showed up there untouched. The redaction only ran over talm's own preview, never over the server diff.

Fix: talm now redacts the `ModeDetails` diff by value before printing, masking both user encrypted-value secrets and the Talos bootstrap material (`ca.key`, `token`, encryption secrets) that the same block was also exposing as context lines. `--show-secrets-in-drift` governs this surface too.

Verified end-to-end against a live node: by default the user secret renders as `value: ***` and the bootstrap keys as `key: ***`; with `--show-secrets-in-drift` the block prints verbatim.

sircthulhu (dismissed):
lgtm

sircthulhu (approved):

PR #224, opened 2026-07-07
fix(charts): disable KVM nested virtualization in Talos presets (CVE-2026-53359)

kvaps:
## Summary

Disables KVM nested virtualization in the Talos machine config produced by the `cozystack` and `generic` presets, to mitigate **CVE-2026-53359** ("Januscape"), a guest-to-host escape in the Linux KVM/x86 shadow MMU (a use-after-free in `kvm_mmu_get_child_sp()` from shadow-page role confusion).

The vulnerable code path is only reachable when a guest is allowed to use nested virtualization; disabling it on the host removes the attack surface regardless of the node kernel patch level. Fixed upstream in `81ccda30b4e8` / stable 6.12.95, 6.6.144, 6.1.177, 6.18.38, 7.1.3 (2026-07-04); no released Talos ships the fixed kernel yet.

## Change

Adds `kvm_intel.nested=0` and `kvm_amd.nested=0` to `machine.install.extraKernelArgs` in both presets. Both args are set so a single config works on Intel and AMD hosts (the non-matching one is silently ignored).

## Notes

- Presets are embedded assets, so this reaches users through a new tagged talm release (and the downstream cozystack bump that pins talm).
- Takes effect on the next reboot (kernel module parameter). Remove these args if a workload legitimately needs nested virt and the node kernel is patched.

Part of a coordinated set of PRs (see Related PRs below).

## Related PRs

Coordinated **CVE-2026-53359** (Januscape) nested-virtualization mitigation across Cozystack repos:

- cozystack/cozystack#3234 — E2E Talos machine-config
- cozystack/talm#224 — Talos `cozystack`/`generic` presets (canonical source)
- cozystack/website#602 — install docs (all versions)
- cozystack/ccp#17 — Talos bootstrap/upgrade skills (prescribe + verify, survives upgrades)
- cozystack/ansible-cozystack#60 — generic Ubuntu/RHEL/SUSE hosts (modprobe.d)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated machine configuration to disable KVM nested virtualization on Intel and AMD by injecting required kernel arguments.
  * Added Talos-version-aware handling for UKI boot settings to prevent incompatible configuration output across templates.
* **Tests**
  * Added contract tests to verify generated Helm config includes the nested-virtualization mitigation and that full merged configuration passes validation without UKI/argument conflicts.

### DIFF
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -165,6 +165,25 @@ machine:
     {{- end }}
     {{- (include "talm.discovered.disks_info" .) | nindent 4 }}
     disk: {{ include "talm.discovered.system_disk_name" . | quote }}
+    {{- /* CVE-2026-53359: disable KVM nested virtualization (guest-to-host
+           escape mitigation). kvm_intel/kvm_amd are built into the Talos
+           kernel, so the nested= parameter takes effect only from the
+           kernel command line — a modprobe.d drop-in or a runtime /sys
+           write cannot set it (the sysfs knob is read-only post-boot).
+           Both modules are listed so one config works on Intel and AMD;
+           the non-matching module silently ignores its arg.
+
+           On Talos >1.11 the generated base config defaults
+           machine.install.grubUseUKICmdline to true (UKI cmdline), and
+           Talos rejects extraKernelArgs alongside it ("install.extraKernelArgs
+           and install.grubUseUKICmdline can't be used together"). Pin it
+           false so the args land on the Talos-built cmdline. The guard
+           matches the base default exactly: emit the field only where the
+           base bundle would set it true, so we never surface a key on a
+           node whose schema predates grubUseUKICmdline. */ -}}
+    {{- if or (not .TalosVersion) (not (semverCompare "<1.12.0-0" .TalosVersion)) }}
+    grubUseUKICmdline: false
+    {{- end }}
     extraKernelArgs:
     # CVE-2026-53359: disable KVM nested virtualization (guest-to-host escape mitigation)
     - kvm_intel.nested=0
--- a/charts/cozystack/templates/_helpers.tpl
+++ b/charts/cozystack/templates/_helpers.tpl
@@ -165,6 +165,10 @@ machine:
     {{- end }}
     {{- (include "talm.discovered.disks_info" .) | nindent 4 }}
     disk: {{ include "talm.discovered.system_disk_name" . | quote }}
+    extraKernelArgs:
+    # CVE-2026-53359: disable KVM nested virtualization (guest-to-host escape mitigation)
+    - kvm_intel.nested=0
+    - kvm_amd.nested=0
 {{- end }}
 
 {{- /* Shared cluster section */ -}}
--- a/charts/generic/templates/_helpers.tpl
+++ b/charts/generic/templates/_helpers.tpl
@@ -58,6 +58,25 @@ machine:
   install:
     {{- (include "talm.discovered.disks_info" .) | nindent 4 }}
     disk: {{ include "talm.discovered.system_disk_name" . | quote }}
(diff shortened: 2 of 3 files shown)


lexfrei:
Heads up — I pushed a fix commit (`1ff68bd`) to this branch. The `extraKernelArgs` approach as-is produces a config Talos rejects on apply for v1.12+.

**Why it failed:** talm's apply path serializes the generated base config merged with the preset. On Talos >1.11 the base defaults `machine.install.grubUseUKICmdline: true` (also when `--talos-version` is empty — it falls back to the current contract), and Talos validation rejects `install.extraKernelArgs` alongside it:

```
install.extraKernelArgs and install.grubUseUKICmdline can't be used together
```

So the kvm args on their own made every v1.12+ apply fail. Confirmed against a live v1.12.6 node — `apply --dry-run` shows the base carrying `grubUseUKICmdline: true`.

**The fix:** pin `grubUseUKICmdline: false` in both presets, gated to exactly where the base sets it true (empty `--talos-version`, or ≥1.12); omitted on explicit <1.12, whose schema predates the field. Added contract tests that render the full apply config and run Talos's own validation, so the conflict can't regress.

Two things worth calling out:

1. **Tradeoff:** this switches v1.12+ nodes from the UKI cmdline back to the Talos-built cmdline. Harmless for standard installs, but relevant if anyone relies on UKI/secure-boot cmdline measurement.

2. **Rollout:** `machine.install.*` kernel args are written by the installer, so this takes effect on `talm upgrade` (installer re-run), not a plain `apply --mode=reboot`. Verified on a live v1.12.6 node: after `apply --mode=reboot` the cmdline was unchanged and `kvm_intel.nested` stayed `Y`; after `upgrade` the cmdline gained `kvm_intel.nested=0 kvm_amd.nested=0` and `nested` flipped to `N`. The "takes effect on the next reboot" line should be corrected, and the coordinated E2E / docs PRs should reflect the upgrade requirement.

kvm_intel/kvm_amd are built into the Talos kernel, so the cmdline is the only way to set `nested=` — a modprobe.d drop-in or a runtime `/sys` write can't (the sysfs knob is read-only post-boot).

kvaps:
Thanks @lexfrei — good catch, and the fix is right. Confirmed the same conflict: `extraKernelArgs` alone is rejected on Talos ≥1.12 because the base defaults `grubUseUKICmdline: true`, and even where it isn't rejected it's a no-op under the UKI cmdline. Pinning `grubUseUKICmdline: false` (gated to where the base sets it true) is the correct fix, and running Talos's own validation in the contract tests is exactly what this needed. The UKI-cmdline → Talos-built-cmdline tradeoff is fine for us — we don't rely on UKI cmdline measurement.

Propagated the same pin to the coordinated PRs, and corrected the rollout wording to "applies on `talm upgrade`, not `apply --reboot`" per your live check:
- cozystack/cozystack#3234 (e2e patch)
- cozystack/ccp#17 (bootstrap overlay + reference docs)
- cozystack/website#602 (install docs, ≥1.12 versions)

Also confirms your point that kvm_intel/kvm_amd are built into the Talos kernel — so the cmdline is the only lever; the ansible-cozystack PR stays on modprobe.d, since on generic distros those are loadable modules.

One open point: the older docs versions (v0–v1.2) pin Talos v1.10.3, where `grubUseUKICmdline` doesn't exist yet but UKI is already the default — so `extraKernelArgs` is a no-op on UEFI there and only Image Factory fixes it. Handling those docs separately.

IvanHunters (approved):
LGTM. The `<1.12.0-0` semver guard correctly emits `grubUseUKICmdline: false` on Talos >=1.12 (including 1.12 prereleases) and omits it on <1.12 where the field doesn't exist; verified against Masterminds/semver behavior and the Talos v1.11.6 / v1.12.6 schemas. The contract test's UKI-conflict assertion matches the actual Talos validation string. Strong tests.

PR #231, opened 2026-07-21
chore(deps): migrate to Helm 4 and drop the cozystack/talos fork

lexfrei:
## What

Migrates talm off the end-of-life Helm 3 library to Helm 4. The v4 dependency graph forces the rest: it drops the `cozystack/talos` fork for stock upstream Talos v1.13.7, then refreshes the dependency tree. Four self-contained commits: golden snapshots (on v3), the v3-to-v4 swap, fork drop plus local `--skip-verify`, remaining dependency bumps.

## Why the fork drop rides along

Helm v4 requires `k8s.io/api` v0.36, which cascades across the whole build graph. The pinned `cozystack/talos` fork was a January-2026 pre-v1.13 snapshot that can't coexist with that bump. The fork only ever existed to carry `--skip-verify` (upstream declined it in siderolabs/talos#12652), so I reimplemented that flag in talm's own client wrappers and moved to upstream v1.13.7. You can't take Helm 4 without also unpinning talos.

## Transparency

The golden snapshots prove the migration doesn't change what talm generates. They were committed on v3, the v4 commit doesn't touch them, and `TestGoldenRender` passes byte-for-byte across the full matrix (cozystack and generic, controlplane and worker, multidoc and legacy). Rendered machine-config is identical before and after.

## Breaking change: `--skip-verify` on wrapped talosctl commands

Under the fork, `--skip-verify` was injected at the library level and worked for every command, including passthroughs (`talm get`, `talm health`, `talm dashboard`). Those run upstream code that never reaches talm's client wrappers, so without the fork there is no way to inject it. `--skip-verify` now works only for talm-native commands (apply, template, upgrade, rotate-ca). Passthroughs print an explicit warning and connect with full verification, instead of failing with an opaque TLS SAN error.

## Testing

`go build`, `go vet`, the full suite, and golangci-lint are all clean. The `--skip-verify` reimplementation covers the TLS config, native routing, the no-nodes contract, config-context and endpoint threading, and the passthrough warning. Manual steps are in `docs/manual-test-plan.md` D4. Reviewed by Opus. Docs for the `--skip-verify` scope reduction will follow at cozystack/website.

## Summary by CodeRabbit

* **New Features**
  * Added reliable `--skip-verify` TLS behavior for talm-native commands (apply, template, upgrade, rotate-ca).
  * Skips **server** certificate verification while preserving client-certificate authentication and talosconfig-based connection targeting.
* **Bug Fixes**
  * Using `--skip-verify` with wrapped talosctl passthrough commands now warns and falls back to full TLS verification.
* **Documentation**
  * Expanded the manual test plan with positive/negative `--skip-verify` scenarios.
* **Tests**
  * Added comprehensive `--skip-verify` routing/TLS coverage and new golden render snapshot validation.


IvanHunters:
## Verdict

LGTM with non-blocking notes.

Helm v3→v4 migration and the cozystack/talos fork drop are mechanically sound and independently verified end-to-end (build, vet, full test suite, lint, `go mod verify`/`tidy` all clean in this environment); only stale-config cleanup items remain.

## Findings

[MINOR] `.github/renovate.json:8-16` — dead Renovate custom manager left behind after the migration it was tracking.

The custom manager's regex (`helm\.sh/helm/v3 v(?<currentValue>\d+\.\d+\.\d+)`) exists specifically to surface the v3→v4 major bump in `go.mod`. This PR completes that bump and removes every `helm.sh/helm/v3` reference from `go.mod` (confirmed: `grep -c "helm.sh/helm/v3" go.mod` → 0). The custom manager can now never match anything again and should be deleted as part of this PR rather than left as orphaned config for the next person to puzzle over.

[MINOR] `.golangci.yml:38-49` — `gomoddirectives.replace-allow-list` entries for the dropped fork are now stale.

The block allow-lists `github.com/siderolabs/talos` and `github.com/siderolabs/talos/pkg/machinery` with a comment justifying it by "the cozystack fork of Talos carries a downstream-only patch ... the replace directive is the only way to consume the fork." This PR removes the `replace (...)` block from `go.mod` entirely — verified `grep -n "^replace\|=> " go.mod` returns nothing, i.e. there are zero `replace` directives left in the module at all. The allow-list and its justifying comment are dead configuration that should be removed alongside the `replace` block it was written for.

[MINOR] `pkg/commands/root.go:71` — a partial client cert/key pair in the talosconfig context is silently ignored.

`skipVerifyTLSConfig` guards with `if configContext.Crt == "" || configContext.Key == ""` and returns a TLS config with no client certificate whenever *either* half is missing. If a context supplies only one of the two (e.g. `Crt` set, `Key` empty — a malformed/half-populated talosconfig), client-certificate authentication is silently dropped with no warning, and the user later hits an opaque authentication failure at connect time rather than a legible error pointing at the bad context. Prefer failing fast (or at least warning) on an XOR-populated pair instead of treating "one half present" the same as "neither present".

[MINOR] Commit messages — the disclosed `--skip-verify` scope-reduction is not marked as a breaking change per Conventional Commits.

The PR body has an explicit "Breaking change: `--skip-verify` on wrapped talosctl commands" section (skip-verify now works only for talm-native commands; wrapped passthroughs like `talm get`/`talm health`/`talm dashboard` silently lost coverage they had under the fork and now warn-and-fall-back instead). None of the three commits use a `!` after `type(scope)` or carry a `BREAKING CHANGE:` footer, despite all being `chore(deps):` — a type conventionally read as safe-to-auto-merge by changelog/semver tooling. Since this repo has no `CONTRIBUTING.md`/`AGENTS.md` mandating the marker, this is a convention nit rather than a hard rule violation, but it is exactly the kind of case the marker exists for: a user-visible CLI behavior loss riding inside a "chore" commit.

## Caveats

- Cozystack chart upgrade/fresh-install/config-matrix checks are not applicable: this PR touches zero packages and zero charts. The only YAML touched is Helm-engine golden-snapshot test fixtures under `pkg/engine/testdata/golden/`, which are test data, not shipped charts.
- Fork-drop behavioral-parity check was executed, not just reasoned: fetched the actual dropped fork commit (`cozystack/talos@d18a185e3680`) and diffed it against upstream Talos v1.12.1/v1.13.7 source. Finding: the fork's own added `client.WithSkipVerify()` option / `Options.skipVerify` field / `getConn()` branch were dead code — never wired into `global.Args.WithClientSkipVerify`, which only ever called `client.WithTLSConfig(tlsConfig)`. That option already short-circuits `getConn()` in stock upstream (verified in v1.12.1 and v1.13.7 source) before the SideroV1/basic-auth/interceptor logic. So the fork's actual observable behavior was already governed by the same code path this PR's reimplementation uses directly — no functional loss found in the connection/TLS layer from dropping the fork.
- Node-context parity for `--skip-verify` (no-nodes vs with-nodes contract across `WithClientNoNodes`/`WithClient`/`withApplyClientBare`/`runTemplate`) was traced call-by-call and cross-checked against the shipped tests (`TestWithClientNoNodes_SkipVerifyOmitsNodeMetadata`, `TestWithClient_SkipVerifySetsResolvedNodes`, `TestWithClient_SkipVerifyPreservesConfigContext`) — all pass and assert the exact contract described in code comments. No stray `GlobalArgs.With*` call sites left bypassing the new `SkipVerify` routing.
- Golden-render byte-identity claim ("rendered machine-config is unchanged") verified independently: `git log --oneline -- pkg/engine/testdata/golden/` shows only the pre-migration commit ever touched those fixtures — none of the three migration commits do — and `go test ./pkg/engine/...` passes `TestGoldenRender` across the full 8-case matrix (cozystack/generic × controlplane/worker × multidoc/legacy).
- `go build ./...`, `go vet ./...`, `go test ./...` (full suite), `golangci-lint run ./...` (0 issues), `go mod verify` (all modules verified), and `go mod tidy` (zero diff against the committed `go.mod`/`go.sum`) were all executed in this run, not assumed from the PR body. One test (`TestCommittedTextFilesIgnoresUntrackedArtefacts`, unrelated to this PR) failed only under an outdated default `git` (2.23.0, missing `git init -b`); re-run with a newer `git` (2.50.1) on `PATH` and the full suite is green. Environment artifact, not a PR defect.
- Deeper behavioral verification of the wider transitive dependency bump (k8s.io 0.35→0.36, fluxcd/pkg/ssa, etcd, cosi-project, etc., forced by Helm v4's `k8s.io/api` v0.36 requirement) is bounded by what the existing test suite exercises (golden render + skip-verify + engine tests); no separate fuzzing of the newly-bumped transitive graph was performed beyond that.

## Recommended follow-ups

- Delete the dead `helm.sh/helm/v3 → v4` custom manager from `.github/renovate.json` and the stale `github.com/siderolabs/talos*` entries from `.golangci.yml`'s `gomoddirectives.replace-allow-list` (both now orphaned by this PR's own changes).

IvanHunters:
On the scope question: the reduced set (`apply`, `template`, `upgrade`, `rotate-ca`) is sufficient, and here's the reasoning, not just a yes.

`--skip-verify` exists for exactly one situation: reaching a node by an IP that is absent from the server cert SANs, while still presenting the client cert (it is not talosctl's `--insecure`, which is maintenance-mode with no client auth at all). That situation is only *functionally required* during the config-management lifecycle: you connect to a not-yet-in-SAN node to push/upgrade the very config that adds the IP to `certSANs`. The four retained commands are precisely those, so the flag keeps working everywhere it is load-bearing.

The passthroughs that lose it (`get`, `health`, `dashboard`, …) run upstream RunE code directly, and upstream declined the flag (siderolabs/talos#12652), so there is no injection point without the fork — and the fork is what blocks Helm 4. Losing it there is a convenience gap in the transient bootstrap window, not a dead end:

- once `apply --skip-verify` lands the config with the IP in `certSANs`, every command (passthroughs included) works with full verification;
- `--endpoints` is preserved, so a node reachable by a SAN-listed name/endpoint is introspectable without skip-verify;
- the new behavior is an explicit warning + fallback to full verification, which is strictly better than the opaque `x509: certificate is valid for … not <ip>` failure you'd otherwise hit.

The only way the scope would be *insufficient* is a supported workflow where you must introspect (`get`/`health`) a node that is *permanently* reachable only by an out-of-SAN IP and cannot be fixed via `apply`. That shouldn't exist in normal usage — `certSANs` should carry the management IPs, and if they don't, `apply` (which keeps the flag) is the fix. So: scope is enough.

IvanHunters (approved):

Issue #230, opened 2026-07-09
cozystack preset: expose machine.time.servers (NTP) via values


lexfrei:
## Problem

There is no knob to set upstream NTP servers, so nodes on isolated networks that must point at a local time source (e.g. the gateway `192.0.2.1`) have to hand-edit `machine.time.servers`.

## Proposal

```yaml
timeServers: [192.0.2.1]
```

rendering `machine.time.servers` when the list is non-empty.

Issue #229, opened 2026-07-09
cozystack preset: custom registry mirrors via values


lexfrei:
## Problem

The preset hardcodes a single mirror (`docker.io` → `https://mirror.gcr.io`) in both the legacy `registries.mirrors` block and the multidoc `RegistryMirrorConfig`. Operators mirroring additional registries (e.g. `ghcr.io`), or pointing at a private pull-through cache with a custom TLS/insecure setting, must fork the preset.

## Proposal

Expose the mirrors via values, driving both the legacy `registries` block and the `RegistryMirrorConfig` documents; keep the current gcr mirror as the default:

```yaml
registryMirrors:
  docker.io: { endpoints: [https://registry.example.com/v2/docker.io] }
  ghcr.io:   { endpoints: [https://registry.example.com/v2/ghcr.io], insecureSkipVerify: true }
```

hiSandog:
Should custom mirrors merge with the preset defaults or replace them entirely? Defining precedence for duplicate registries and whether authentication/TLS settings are supported alongside mirror endpoints would keep the values contract predictable for air-gapped installations.

Issue #228, opened 2026-07-09
cozystack preset: passthrough extraArgs for apiServer/controllerManager/scheduler


lexfrei:
## Problem

The `cozystack` preset templates a fixed set of control-plane component args (`oidc-*`, `bind-address`, `allocate-node-cidrs`). Any other arg — for example `max-requests-inflight` / `max-mutating-requests-inflight` on kube-apiserver — requires forking the preset. `kubelet` already has `extraKubeletExtraArgs`; the control-plane components have no equivalent.

## Proposal

Add passthrough maps mirroring `extraKubeletExtraArgs`, merged into each component's `extraArgs`, with the same collision guard the sysctls/kubelet knobs use:

```yaml
extraApiServerArgs: { max-requests-inflight: "2000", max-mutating-requests-inflight: "1000" }
extraControllerManagerArgs: {}
extraSchedulerArgs: {}
```

Issue #227, opened 2026-07-09
cozystack preset: support multiple Layer2 VIPs (VIP on more than one link)


lexfrei:
## Problem

`vipLink` is a single value, so the preset can pin at most one VIP. A node that needs a VIP on **two** links — e.g. a control-plane VIP on the primary link (`192.0.2.254`) and a separate VIP on a storage/secondary link (`203.0.113.254`) — cannot express the second and must hand-edit the rendered config.

Related to #226 (full multi-link topology).

## Proposal

Accept a list of VIPs, each bound to a link:

```yaml
vips:
  - { link: bond0, ip: 192.0.2.254 }
  - { link: bond1, ip: 203.0.113.254 }
```

mapping to one `Layer2VIPConfig` (Talos 1.12+) or an inline `vip` block (legacy) per link. Keep `vipLink`/`floatingIP` as the single-VIP shorthand for back-compat.

PR #232, opened 2026-07-21
feat(charts): add preset value knobs (#226-#230)

lexfrei:
New values so a node's Talos config can come from `values.yaml` instead of a template fork:

- `timeServers` sets `machine.time.servers`
- `extraApiServerArgs` / `extraControllerManagerArgs` / `extraSchedulerArgs` pass through control-plane component args
- `registryMirrors` declares machine registry mirrors
- `registryTLS` sets TLS posture per registry endpoint host (a private CA, or skipping verification for a self-signed cache)
- `vips` gives multiple Layer2 VIPs, each pinned to a link
- `network.preserveExisting` keeps the running interfaces verbatim
- `network.extraLinks` declares bonds, VLANs, and extra addresses, with bond tuning, mtu and routes

Covers #226, #227, #228, #229, #230.

Every knob defaults to empty, so a stock render is byte-identical to before. `TestGoldenRender` pins that across both presets, control-plane and worker, legacy and multi-doc. Nothing changes for existing users until they opt in.

Anything Talos would reject fails fast at render with a hinted message instead of shipping a broken config: an `extraArgs` key colliding with a preset built-in, a duplicate or dangling VIP, a VIP on a link already in the running interfaces, a VLAN on a missing parent, an empty or mode-less bond, a route without a gateway, a malformed registry entry. Component args are `map[string]string` on the Talos side, so values are quoted to keep an unquoted numeric from becoming a YAML int.

The apply gate learned that a link this config creates (a bond or VLAN from `extraLinks`) satisfies references to it from sibling documents, so an extraLinks bond with a VLAN and VIP on top passes on first apply instead of blocking.

Contract tests cover the accepted and rejected path for every knob, on both schemas and both presets. `docs/manual-test-plan.md` gets a matching section.

One follow-up, not in this PR: operator docs on cozystack/website for the new values.

## Summary by CodeRabbit

* **New Features**
  * Added chart knobs: `timeServers`, component `extra*Args`, `registryMirrors`, and `registryTLS`.
  * Added multi-VIP support via `vips`, and improved Talos v1.12+ multi-document network output.
  * Added `network.preserveExisting` and `network.extraLinks` for richer declarative topology (including bonds/VLANs/addresses/routes).

* **Bug Fixes**
  * Stricter VIP/IP canonicalization and collision checks; VIPs are removed from discovered link addresses.
  * Apply-time reference validation now accounts for links created within the same rendered apply (with WireGuard-specific behavior).

* **Documentation**
  * Updated README and manual test plan for the new knobs and safety gates.

* **Tests**
  * Expanded contract and apply validation coverage for legacy and multidoc paths.


kvaps (dismissed):
Reviewed by checking out the branch and verifying each claim in the description independently rather than relying on the added tests.

## Verification

**"a stock render is byte-identical to before" — confirmed, with room to spare.** I rendered a 240-cell matrix (3 discovery fixtures × 2 charts × 2 templates × 4 Talos versions × 5 value sets) on both the merge base and the branch head and diffed them: identical. The fixtures included a rich topology (physical primary + bond with two slaves + VLAN child + bridge with a port, a link-local address, a non-default route, jumbo MTU) and the value sets covered `floatingIP`, `vipLink`, `oidcIssuerUrl` and `allocateNodeCIDRs: false`. The committed golden files are untouched and `TestGoldenRender` passes across both presets × controlplane/worker × legacy/multi-doc. Moving the hardcoded `docker.io` mirror into `values.yaml` is handled carefully, and the deep-merge caveat (`{}` keeps the default, `null` clears it) is both documented and true — I checked that `registryMirrors: null` emits no registry documents at all.

**Component-arg quoting — confirmed.** Integers, floats and booleans all come out quoted; legitimate values survive intact (`A=true,B=false` stays whole, an empty string does not become `null`, `yes` does not become a boolean). Key ordering is deterministic.

**Apply-gate relaxation — confirmed as bounded.** I exercised `WalkRefs` + `ValidateRefs` directly. A bond → VLAN → VIP chain created entirely by the config validates clean, while a typoed `VLANConfig.parent`, a dangling `Layer2VIPConfig.link`, a bond slave absent from the node, and a VIP pointing at a near-miss VLAN name all still block. Because the created-link union runs in its own pass before validation, document order does not matter. `LinkConfig` still validates its own name, and the `WireguardConfig` exception is described accurately in the README.

**"Anything Talos would reject fails fast" — mostly.** Every case listed in the description does fail with a clear message on both schemas, and each has a contract test. A few gaps are below.

`go build`, `go vet`, `gofmt -l`, `go test ./...` and `golangci-lint run` are all clean. The 97 new contract tests assert on literal rendered output rather than restating the template expression, and cover the negative path for every guard — no skips, no TODOs.

## Please fix

**`registryTLS` keys are emitted unquoted, unlike `registryMirrors`.** `charts/talm/templates/_helpers.tpl:366` (multi-doc) and `:399` (legacy) interpolate the key raw, while the mirror name goes through `| toYaml` at `:356` and `:389` for exactly this reason. An IPv6 endpoint host produces invalid YAML on both schemas:

```yaml
# registryTLS: { "[::1]:5000": { insecureSkipVerify: true } }
kind: RegistryTLSConfig
name: [::1]:5000          # yaml: did not find expected node content
```

```yaml
      [::1]:5000:          # same failure in machine.registries.config
        tls:
```

Neither document parses. A DNS `host:port` key survives only because it happens to contain no `": "`. The same `| toYaml` treatment in both places fixes it.

## Worth considering

**An `extraLinks` bond can enslave a link the chart is also emitting a config for.** At `charts/talm/templates/_helpers.tpl:1375-1400`, `bond.interfaces` are not checked against `$configurableLinks`, `$existingLinkNames` or `$emittedNames` — only the entry's own `interface` name is. Given a node whose discovered primary is `eth0`:

```yaml
network:
  extraLinks:
    - interface: bond1
      bond: { interfaces: [eth0], mode: 802.3ad }
      addresses: [203.0.113.10/24]
```

the render emits both:

```yaml
kind: LinkConfig
name: eth0
addresses:
  - address: 192.168.201.10/24
routes:
  - gateway: 192.168.201.1
---
kind: BondConfig
name: bond1
links:
  - eth0
```

The discovery path deliberately filters this shape out — `configurable_link_names` (`:736-747`) excludes anything with a `slaveKind` so a slave never gets a standalone document alongside its master. The declarative path has no equivalent check, and "move the primary NIC into a bond" is one of the main reasons to reach for `extraLinks`. The same applies under `preserveExisting`: a preserved interface emitted verbatim in `machine.network.interfaces` can simultaneously become a bond slave. Given how thorough the rest of the guard set is, this reads more like an oversight than a decision — but if it is intentional, a note in the `extraLinks` values documentation would help.

**Prefix length is not validated in address and route guards.** `talm.guard.addresses_not_vip` (`:428-434`) and `talm.render.link_routes_mtu` (`:452-459`) check for the presence of `/` and validate the IP portion, but not the prefix itself. `203.0.113.10/33`, `203.0.113.10/abc` and `203.0.113.10/` all render successfully and reach the node, where Talos fails at decode with a message that does not name the document — the exact failure mode these guards exist to prevent. `cidrPrefixLen` is already registered and used elsewhere in the same file (`:644`, `:779`) and returns `-1` on a parse failure, so `ge (cidrPrefixLen $address) 0` would close both.

**`mtu` is not validated at all** (`:470-472`). `mtu: "jumbo"`, `mtu: -5` and `mtu: 99999` are emitted verbatim; `mtu: 0` is silently dropped as falsy.

**Non-scalar `extraArgs` values leak a Go string.** The guard catches only `kindIs "invalid"`, so a nested map renders as `foo: "map[a:b]"` onto the component command line. Affects `charts/cozystack/templates/_helpers.tpl:238,254,277` and `charts/generic/templates/_helpers.tpl:110,120,130`.

## Nits

- `pkg/applycheck/refs.go:249-252` — the `handleListOnly` docstring still says it emits "only the list-valued slaves/ports of the doc, not the doc's own `.name`", but the function now calls `appendCreatedRef` first. Stale by exactly the change this PR makes.
- `charts/talm/templates/_helpers.tpl:610-613` — the comment justifies the dict-mutation pattern by stating that `range` introduces a new scope so `$var = ...` does not propagate. In Go `text/template`, `=` assigns to the existing variable and does survive iterations; the same file relies on that at `:781`, `:1005` and `:1374`. Either drop the rationale or bring `$best` in line with the surrounding style.
- Duplicate addresses across two `extraLinks` entries, and an `extraLinks` address matching another link's discovered address, are not caught. Probably harmless, but they sit just outside the "anything Talos would reject" line.
- `Layer2VIPConfig` for a VIP on an `extraLinks` bond is emitted before the `BondConfig` that creates it. Documents are a set, so this is fine today.
- `timeServers` accepts a map and renders it as-is under `servers:`.

## Size

The +3105 is 1936 lines of tests plus 156 of docs; the logic itself is around 740 lines. Splitting would create artificial ordering between PRs since the guards are shared across knobs (`$vipIPs` is needed by both `vips` and `extraLinks`; `talm.guard.registries` serves both schemas), so I would keep it as one change.

charts/talm/templates/_helpers.tpl
```diff
@@ -270,19 +270,223 @@ busPath: {{ .spec.busPath }}
...
+endpoints:
+{{- range $cfg.endpoints }}
+  - url: {{ . }}
+{{- end }}
+{{- end }}
+{{- range $host, $cfg := .Values.registryTLS }}
+---
+apiVersion: v1alpha1
+kind: RegistryTLSConfig
+name: {{ $host }}
```
myasnikovdaniil: **B1 (blocker).** A bracketed IPv6 endpoint host renders as `name: [fd00::1]:5000`, and `[` opens a YAML flow sequence — the document no longer parses (`yaml: did not find expected key`). Verified by rendering `registryTLS: {"[fd00::1]:5000": {insecureSkipVerify: true}}` and feeding the output to a YAML decoder.

`registryMirrors` at line 354 already handles this with `name: {{ $name | toYaml }}` and a comment about the `*` fallback key. Same treatment needed here: `{{ $host | toYaml }}`.

(Plain `host:port` without brackets parses fine, so only the bracketed form is affected.)

charts/talm/templates/_helpers.tpl
```diff
@@ -270,19 +270,223 @@ busPath: {{ .spec.busPath }}
...
+        endpoints:
+{{- range $cfg.endpoints }}
+        - {{ . }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- with .Values.registryTLS }}
+    config:
+{{- range $host, $cfg := . }}
+      {{ $host }}:
```
myasnikovdaniil: **B1 (blocker), legacy half.** Same unquoted `$host` as line 366, this time as a mapping key:

```
    config:
      [fd00::1]:5000:
        tls:
          insecureSkipVerify: true
```

Parsing this fails with `yaml: line 65: did not find expected key`. `{{ $host | toYaml }}` fixes it here too.

charts/talm/templates/_helpers.tpl
```diff
@@ -798,14 +1062,20 @@ link: {{ .Values.vipLink }}
...
-{{- $addresses = append $addresses . }}
+{{- /* Strip any VIP ip ($vipIPs, built from the coerced floatingIP
+       plus the vips list at the top of the define) from this link's
+       addresses. Going through `printf "%s/" $.Values.floatingIP`
+       directly would emit `%!s(int=192168)/` for a numeric YAML
+       scalar on a worker render, so the coerced values are used. */ -}}
+{{- $addr := . }}
+{{- $isVip := false }}
+{{- range $vipIPs }}
+{{- if hasPrefix (printf "%s/" .) $addr }}
```
myasnikovdaniil: **B2 (blocker).** This compares raw strings, but Talos reports discovered addresses canonically via `netip.Prefix.String()`. A VIP spelled `2001:0DB8::5` therefore never matches the discovered `2001:db8::5/64`, so the strip misses it.

Verified: fixture with `2001:db8::5/64` on eth0 plus `vips: [{link: eth0, ip: "2001:0DB8::5"}]` renders the address **both** under `LinkConfig.addresses` and as a `Layer2VIPConfig`. A control render using the canonical spelling strips correctly, so the mechanism works — only the comparison is spelling-sensitive.

That is the exact condition the comment above warns about ("leaves the leader and the followers disagreeing about who owns it"). The same gap pre-exists for `floatingIP` on main (lines 164 and 807), so canonicalizing once — before the dedup check, the `addresses_not_vip` guard, and this strip — covers both values.

charts/talm/templates/_helpers.tpl
```diff
@@ -270,19 +270,223 @@ busPath: {{ .spec.busPath }}
...
+{{- range .addresses }}
+{{- /* Talos decodes these into netip.Prefix, so a bare IP fails at DECODE
+       time ("no '/'") — before validation, with a message that does not
+       name the document. The discovery path filters malformed CIDRs; the
+       declarative path refuses them. */}}
+{{- if not (contains "/" (. | toString)) }}
+{{- fail (printf "talm: address %q on network.extraLinks %q has no prefix length. Talos parses addresses as CIDR, so write it as 203.0.113.10/24." (. | toString) $name) }}
+{{- end }}
+{{- $ip := (splitList "/" (. | toString)) | first }}
+{{- if not (ipIsValid $ip) }}
```
myasnikovdaniil, line 478: `ipIsValid` runs on the text before the slash only, so the prefix length is never checked. `203.0.113.10/999` renders straight through to the emitted document with no error — Talos then rejects it at decode time with a message that does not name this document, which is the outcome the surrounding guards exist to avoid.

`cidrPrefixLen` already exists (`pkg/engine/helm/engine.go:265`) and validates the whole CIDR. The route-destination check at line 457 has the same partial-validation shape (`198.51.100.0/foo` also passes).

charts/talm/templates/_helpers.tpl
```diff
@@ -1016,4 +1287,194 @@ name: {{ $fipStr | quote }}
...
+{{- end }}
+{{- range $vlan := $extra.vlans }}
+{{- if not $vlan.vlanId }}
+{{- /* VLANConfig requires vlanID on the wire; without it the emitted
+       document has name "<iface>." and vlanID: <nil>, which Talos
+       rejects. Fail fast with a precise message. */}}
+{{- fail (printf "talm: network.extraLinks entry %q declares a VLAN with no vlanId. VLANConfig requires vlanID; add vlanId to the vlans entry." ($extra.interface | toString)) }}
+{{- end }}
+{{- if or (lt ($vlan.vlanId | int) 1) (gt ($vlan.vlanId | int) 4094) }}
+{{- fail (printf "talm: network.extraLinks entry %q declares vlanId %v, outside the valid range. Talos requires vlanID between 1 and 4094." ($extra.interface | toString) $vlan.vlanId) }}
```
myasnikovdaniil, line 1631: `$vlan.vlanId | int` truncates, so `vlanId: 7.5` becomes `7` and passes this range check — but line 1469 emits the original value. The render output is `vlanID: 7.5` with document name `eth0.7.5`, which Talos rejects while decoding. Worth asserting the value is an integer before range-checking it.

pkg/applycheck/refs.go
```diff
@@ -193,10 +200,15 @@ var multidocHandlers = map[string]multidocHandler{
...
-	// dispatch table so they don't get a name-based Phase 1 finding.
+	// DummyLinkConfig / LinkAliasConfig describe virtual links being
+	// created. Their .name is the new resource, not an existing-link
+	// reference, so it is recorded as created rather than validated —
+	// that way a VLAN or VIP pointing at one of them in the same config
+	// resolves. WireguardConfig also creates a link but is deliberately
+	// absent here: it belongs to the parallel net-addr walker, and the
+	// two dispatch maps must stay disjoint or a kind gets double-walked.
+	"DummyLinkConfig":  handleCreatorOnly,
+	"LinkAliasConfig":  handleCreatorOnly,
```
myasnikovdaniil, line 211: `LinkAliasConfig.name` supports a format verb: machinery v1.13.7 `config/types/network/link_alias.go:59-63` documents `net%d` expanding to `net0`, `net1`, … one per matched link, with the selector allowed to match multiple links in that case.

So `appendCreatedRef` registers the literal `net%d` into `linkSet` while the names that will actually exist are never registered. Net effect is inverted from the intent: a `VLANConfig.parent: net%d` reference passes the gate even though no such link will ever exist, and a legitimate `parent: net0` still blocks on first apply.

This also makes the README's new sentence — that `WireguardConfig` is the one exception which does not register its name — narrower than stated.

myasnikovdaniil (changes requested):
NOT LGTM — two input-dependent rendering paths in the new knobs emit a config that is either unparseable YAML or double-declares a VIP.

**Business context**: Five preset knobs (#226–#230) so a node's Talos config comes from `values.yaml` instead of a forked template — NTP, control-plane `extraArgs`, registry mirrors/TLS, multiple VIPs, and declarative network topology.

## Blockers

### B1: `registryTLS` host rendered unquoted → invalid YAML

**File**: `charts/talm/templates/_helpers.tpl:366` (multidoc), `:399` (legacy)

**Issue**: A bracketed IPv6 endpoint host emits `name: [fd00::1]:5000`, where `[` opens a YAML flow sequence.

**Evidence**: Rendered both schemas with `registryTLS: {"[fd00::1]:5000": {insecureSkipVerify: true}}` and parsed the result:

```
kind: RegistryTLSConfig
name: [fd00::1]:5000        -> yaml: line 99: did not find expected key
```

```
    config:
      [fd00::1]:5000:       -> yaml: line 65: did not find expected key
        tls:
          insecureSkipVerify: true
```

The sibling `registryMirrors` path already solved exactly this at `:354` with `name: {{ $name | toYaml }}`, and its comment explains why. `registryTLS` does not. A plain `host:port` (no brackets) parses fine — only the bracketed form breaks.

**Impact**: `talm template` emits a document that neither talm nor the node can parse — worse than the fail-fast this PR promises elsewhere.

**Fix**: `{{ $host | toYaml }}` at both sites, matching `:354`.

### B2: Non-canonical IPv6 VIP emitted as both a static address and a `Layer2VIPConfig`

**File**: `charts/talm/templates/_helpers.tpl:1073`

**Issue**: The strip compares `hasPrefix (printf "%s/" .) $addr` on raw strings. Talos reports addresses canonically (`netip.Prefix.String()`), so a VIP spelled `2001:0DB8::5` never matches the discovered `2001:db8::5/64`.

**Evidence**: With a fixture carrying `2001:db8::5/64` on eth0 and `vips: [{link: eth0, ip: "2001:0DB8::5"}]`, the render contains **both** `2001:db8::5/64` under `LinkConfig.addresses` and `kind: Layer2VIPConfig`. A control render with the canonical spelling strips correctly, so the mechanism itself works — only the string comparison is wrong. The same gap pre-exists for `floatingIP` on main (`:164`, `:807`); this PR rewrites that loop and extends it to `vips`, so one fix covers both.

**Impact**: Exactly the failure the code's own comment says the strip prevents — "leaves the leader and the followers disagreeing about who owns it". Breaks VIP failover.

**Fix**: Canonicalize every VIP (and the discovered address) once, before the dedup check, the `addresses_not_vip` guard, and this strip.

## Non-blocking follow-ups

1. **CIDR prefix length unvalidated** — `_helpers.tpl:432` and `:457` check `contains "/"` and then run `ipIsValid` on the text *before* the slash only. `203.0.113.10/999` as an `extraLinks` address, and `198.51.100.0/foo` as a route destination, both render straight through with no error. The `cidrPrefixLen` helper already exists (`pkg/engine/helm/engine.go:265`) and would close both paths.

2. **Fractional `vlanId`** — `_helpers.tpl:1449` range-checks `$vlan.vlanId | int`, truncating `7.5` to `7` and passing, but `:1469` emits the original value: the rendered output is `vlanID: 7.5` with document name `eth0.7.5`. Talos rejects that at decode time. Worth checking the value is an integer before the range check.

3. **`LinkAliasConfig` `%d` aliases** — `pkg/applycheck/refs.go:211` registers `.name` literally. Machinery v1.13.7 `config/types/network/link_alias.go:59-63` documents `net%d` expanding to `net0`, `net1`, … per matched link. So the literal `net%d` enters `linkSet` while the names that will actually exist do not: a `net%d` reference passes the gate, and a legitimate `net0` reference still blocks. This also makes the README's new claim that `WireguardConfig` is the sole exemption slightly overstated.

4. **Bare `apiServer:`** — `charts/generic/templates/_helpers.tpl:123` emits a valueless key when `certSANs` is `[]` (the default) and `extraApiServerArgs` is empty. This is pre-existing (`apiServer:` is unconditional on main at line 76), so it is not a regression from this PR and the golden tests correctly show defaults unchanged — but the block is open here and it is a one-line fix.

5. **Commit granularity** — five independent features land in one 3,105-line commit. Split per issue (#226–#230) would make `git bisect` and a per-feature revert possible.

6. **Operator docs** — the PR notes site docs as a follow-up; there is no open PR on the docs repo yet. Worth a tracking issue so the values reference (legacy vs v1.12+ split for `vips`/`registryTLS`, the multi-doc-only restriction on `network.extraLinks`, and the `preserveExisting` escape hatch) does not get lost.

## Bot findings triaged

- Empty `apiServer:` breaking byte-identical defaults — **not a regression**; `apiServer:` is already unconditional on main at line 76. Retained as follow-up 4 on its cosmetic merits only.
- `docs/manual-test-plan.md:646` setup/error-message mismatch — valid.
- README VIP-stripping wording contradicting itself — valid, and B2 makes the wording moot until fixed.
- README MD028 blank lines inside the blockquote — valid, lint-only.

## Verified as sound

`go test ./...` passes locally across all packages, matching CI. Golden renders confirm defaults are byte-identical over both presets × controlplane/worker × legacy/multi-doc. The two applycheck dispatch maps are genuinely disjoint — `WireguardConfig` appears only in `multidocNetAddrHandlers`, as the comment claims. `mergeMaps` (`pkg/engine/engine.go:1795`) does assign `nil`, so the "null the key out" instruction in `values.yaml` is accurate. The 228 test functions across the five contract files do cover accept *and* reject paths; B1 and B2 are cases those tests never reach, not a testing-convention gap.

charts/talm/templates/_helpers.tpl
```diff
@@ -782,9 +1134,52 @@ link: {{ .Values.vipLink }}
...
+{{- range $slave := (($extra.bond | default dict).interfaces | default list) }}
+{{- $slaveName := $slave | toString }}
+{{- if has $slaveName $existingLinkNames }}
+{{- fail (printf "talm: network.extraLinks bond %q enslaves %q, which the running machine.network.interfaces block already declares verbatim (network.preserveExisting). The device would be declared twice from two sources. Drop it from the preserved interfaces, or build the bond from links that block does not name." ($extra.interface | toString) $slaveName) }}
+{{- end }}
+{{- if has $slaveName $configurableLinks }}
+{{- $slaveAddresses := fromJsonArray (include "talm.discovered.addresses_by_link" $slaveName) }}
+{{- if or $slaveAddresses (eq $slaveName $defaultLinkName) }}
+{{- fail (printf "talm: network.extraLinks bond %q enslaves %q, which discovery reports carrying its own addresses or the default route. A bond slave holds no addressing of its own, so this apply would take the node's connectivity away without saying where it goes. Move those addresses and routes onto the bond entry, or enslave a link that carries neither." ($extra.interface | toString) $slaveName) }}
```
myasnikovdaniil: **B3 (blocker).** The message offers "Move those addresses and routes onto the bond entry" as a way out, but the condition on line 1167 tests only `$slaveAddresses` and `$defaultLinkName` — both from discovery. It never reads `$extra.addresses`, so that remediation cannot unblock the render.

Verified: eth0 carrying `192.168.201.10/24` plus the default route via `192.168.201.1`, and an `extraLinks` entry declaring `bond0` over `[eth0, eth1]` with that same address and gateway moved onto the bond entry — the render still fails with this exact message. Bonding an idle NIC works, so the rest of the guard behaves as designed.

The comment above states the intent ("They have to say so by putting the addresses on the bond"), so the code is diverging from the stated design. This currently makes enslaving an already-addressed NIC unreachable — the `bond0`-over-two-NICs-carrying-the-default-route case from #226.

Either honour the intent (accept the case when the bond declares addresses), or drop the first clause so the message only offers the remediation that works.

pkg/engine/contract_network_multidoc_test.go
```diff
@@ -1503,3 +1503,1150 @@ func TestContract_NetworkMultidoc_FloatingIPNotInDiscoveredSubnetFallsBackToGate
...
+	}
+	// The primary link keeps its document.
+	assertContains(t, out, "name: eth0")
+}
+
+// Contract: enslaving a link that carries addresses or the default route
+// fails instead of silently dropping them. Filtering the slave's document
+// would take the node's addressing away without saying where it went, and
+// the render cannot know whether the operator meant to move it.
+func TestContract_NetworkMultidoc_ExtraLinksBondSlaveWithAddresses_Fails(t *testing.T) {
```
myasnikovdaniil, line 2605: This test declares `addresses: [203.0.113.10/24]` on the bond — following the remediation the guard's error message prescribes — and asserts that it still fails. So the contract being pinned here is the misleading guidance rather than the genuine restriction.

Worth pairing with a positive case: slave addresses moved onto the bond entry, render succeeds. That test would have surfaced the mismatch, and it is the coverage that decides which way B3 gets fixed.

myasnikovdaniil (changes requested):
NOT LGTM — both original blockers are fixed; one new blocker came in with the fix round.

## Previous findings — all resolved

- **`registryTLS` host unquoted** — `{{ $host | toYaml }}` now applied at `_helpers.tpl:405` and `:442`. Verified: a bracketed IPv6 endpoint host renders as `name: '[fd00::1]:5000'` and both schemas parse.
- **Non-canonical IPv6 VIP** — the new `ipCanonical` template helper normalises both sides. Verified: `vips[].ip: 2001:0DB8::5` against a discovered `2001:db8::5/64` now strips correctly and emits a single `Layer2VIPConfig`; two spellings of one address are also caught as a duplicate.
- **CIDR prefix length** — `cidrPrefixLen < 0` now checked on `extraLinks` addresses and on route destinations. `203.0.113.10/999` and `198.51.100.0/foo` are both refused, with the document named.
- **Fractional `vlanId`** — digits-only check now runs before the range check; `vlanId: 7.5` is refused.
- **`LinkAliasConfig` `%d` patterns** — `unionCreatedLinks` contributes the pattern's prefix and admits references matching prefix+digits. `aliasPatternPrefix` mirrors machinery's own `Validate` exactly (`strings.Cut` on `%`, rejecting `suffix != "d"` and an empty prefix), so the accepted shape matches what the node will accept.
- **Bare `apiServer:`** — declining this with a rationale comment and a pinning test is a fine answer for a cosmetic pre-existing artifact.

The additional hardening in this round (mtu digits + kernel 68-65535 bounds, the `timeServers` shape guard, and folding six copied `extraArgs` loops into `talm.render.component_args` with map/slice rejection) all render and refuse as intended in local checks.

## Blocker

### B3: bond-slave guard prescribes a remediation that cannot work

**File**: `charts/talm/templates/_helpers.tpl:1167-1168`

**Issue**: The guard refuses to enslave a link that discovery reports carrying addresses or the default route, and its message offers two ways out:

> Move those addresses and routes onto the bond entry, or enslave a link that carries neither.

The condition at `:1167` tests only `$slaveAddresses` and `$defaultLinkName` — both derived from discovery. It never reads `$extra.addresses`. The first remediation therefore has no effect.

**Evidence**: With eth0 carrying `192.168.201.10/24` and the default route via `192.168.201.1`, and an `extraLinks` entry declaring `bond0` over `[eth0, eth1]` with **that same address and gateway moved onto the bond entry**, the render still fails with the identical message. Bonding an idle NIC (no addresses, not the default-route link) succeeds, so the guard is otherwise working as designed.

The guard's own comment states the intended contract — "They have to say so by putting the addresses on the bond" — so this is the implementation diverging from the stated design rather than a deliberate restriction.

`TestContract_NetworkMultidoc_ExtraLinksBondSlaveWithAddresses_Fails` (`pkg/engine/contract_network_multidoc_test.go:2605`) declares `addresses` on the bond, i.e. follows the prescribed remediation, and asserts failure — so the suite currently pins the misleading guidance instead of catching it.

**Impact**: enslaving an already-addressed NIC is unreachable through `network.extraLinks`. That is the headline scenario in #226 — `bond0` as 802.3ad over two NICs carrying the default route — so the most common bonding migration cannot be expressed, and the error text sends the operator down a path that does not resolve it.

**Fix**: either accept the case when the bond entry declares addresses (matches the comment's stated intent), or drop the first clause from the message and narrow the test to pin only the genuine restriction. A test covering the accepted path — slave addresses moved onto the bond, render succeeds — would close the gap either way.

charts/talm/templates/_helpers.tpl
```diff
@@ -782,9 +1134,60 @@ link: {{ .Values.vipLink }}
...
+{{- $slaveName := $slave | toString }}
+{{- if has $slaveName $existingLinkNames }}
+{{- fail (printf "talm: network.extraLinks bond %q enslaves %q, which the running machine.network.interfaces block already declares verbatim (network.preserveExisting). The device would be declared twice from two sources. Drop it from the preserved interfaces, or build the bond from links that block does not name." ($extra.interface | toString) $slaveName) }}
+{{- end }}
+{{- if has $slaveName $configurableLinks }}
+{{- $slaveAddresses := fromJsonArray (include "talm.discovered.addresses_by_link" $slaveName) }}
+{{- if and $slaveAddresses (not $extra.addresses) }}
+{{- fail (printf "talm: network.extraLinks bond %q enslaves %q, which discovery reports carrying its own addresses (%s). A bond slave holds no addressing of its own, so this apply would take the node's connectivity away without saying where it goes. Put those addresses on the bond entry, or enslave a link that carries none." ($extra.interface | toString) $slaveName (join ", " $slaveAddresses)) }}
+{{- end }}
+{{- if and (eq $slaveName $defaultLinkName) (not $extra.routes) }}
```
myasnikovdaniil: Follow-up, non-blocking. This tests only that `$extra.routes` is non-empty, so a route that is not a default route satisfies it.

With `routes: [{gateway: 192.168.201.1, destination: 172.16.0.0/12}]` on a bond over the default-route link, the render is accepted and the emitted `BondConfig` is:

```yaml
routes:
  - gateway: 192.168.201.1
    destination: 172.16.0.0/12
```

No default route anywhere — the node comes back up without a way off its subnet, which is the outcome the message below is written to prevent.

Requiring at least one entry with no `destination` would close it. Not blocking: it takes an operator actively writing a destination while omitting the default, and the render faithfully emits what it was given.

charts/talm/templates/_helpers.tpl
```diff
@@ -782,9 +1134,60 @@ link: {{ .Values.vipLink }}
...
+{{- /* skip the typed per-link rebuild only when preserveExisting
+       is set AND discovery actually found a legacy interfaces block to
+       carry verbatim. A fresh v1.12 node with preserveExisting set has
+       nothing to preserve, so it must fall through to the normal rebuild
+       rather than emit a network-less config. */}}
+{{- if not (and $preserveExisting $legacyInterfaces) }}
 {{- range $linkName := $configurableLinks }}
 {{- $link := lookup "links" "" $linkName }}
-{{- if $link }}
+{{- if and $link (not (has $linkName $bondSlaveNames)) }}
```
myasnikovdaniil, line 1218: Follow-up, non-blocking. Filtering the slave here also drops any non-default static route it carried, and the guard above only inspects addresses and the default route.

Concretely: a link carrying `10.0.0.0/24 via 10.0.0.1` on its own document loses that route when enslaved, with no refusal and no warning. Less severe than losing connectivity, and arguably outside this guard's stated scope — but it is the same class of silent loss, so either extending the check to scoped routes or naming the limitation in the `network.extraLinks` values documentation would round it out.

myasnikovdaniil (dismissed):
LGTM — every blocker from the previous rounds is resolved and independently verified; two narrow edges in the new bond-slave guard are left as follow-ups.

## B3 resolved

The guard is now two independent checks (`_helpers.tpl:1168` and `:1171`) instead of one, and both messages interpolate the discovered values so the operator is told which addresses and which gateway to restate. `talm.discovered.gateway_by_link`, used by the second message, is the pre-existing helper at `:870`.

Verified against a fixture where eth0 carries `192.168.201.10/24` plus the default route via `192.168.201.1`:

- Declaring `bond0` over `[eth0, eth1]` with that address and gateway restated on the bond entry now renders, and neither slave gets a `LinkConfig` of its own.
- Omitting them still refuses, now naming `carrying its own addresses (192.168.201.10/24)`.
- Bonding an idle NIC continues to work.

`TestContract_NetworkMultidoc_ExtraLinksBondSlaveWithAddresses_Fails` now enslaves eth1 — which carries `10.0.0.5/24` and is not the default-route link — so it exercises the addresses branch rather than pinning the old contradiction. The two added tests cover the accepted path and the forgot-the-route case, and the manual-test-plan section flags the bond migration as worth applying on real hardware, which is the right call given the render cannot prove the node keeps its route across the switch.

Local `go test ./...` is clean, including the golden renders — stock output is still byte-identical to before for both presets across controlplane/worker and legacy/multi-doc.

## Non-blocking follow-ups

1. **The default-route check can be satisfied without a default route** (`_helpers.tpl:1171`). The condition tests only that `$extra.routes` is non-empty, so a destination-scoped route passes it. With `routes: [{gateway: 192.168.201.1, destination: 172.16.0.0/12}]` on a bond over the default-route link, the render is accepted and the emitted `BondConfig` carries only that scoped route — no default route anywhere. The guard's purpose is "so the node keeps its way out", and this satisfies the check without achieving it. Requiring at least one route with no `destination` would close it. Non-blocking because it takes an operator actively writing a destination while omitting the default; the render does emit exactly what it was given.

2. **A slave's non-default static route is dropped without a word.** A link carrying a scoped route (e.g. `10.0.0.0/24 via 10.0.0.1`) has it on its own document; enslaving that link drops the document and the route, and the guard does not look at anything beyond addresses and the default route. Lower severity than losing connectivity, and arguably outside the guard's stated scope, but it is the same class of silent loss the guard exists to prevent — worth either extending the check or naming the limitation in the values documentation.

## Earlier findings, all confirmed fixed

`registryTLS` host quoting on both schemas; IPv6 VIP canonicalisation via the new `ipCanonical` helper, covering the strip, the duplicate check and the extraLinks guard; full-CIDR prefix validation on addresses and route destinations; the integer check on `vlanId` ahead of the range check; and `LinkAliasConfig` `%d` patterns registering as a prefix so `net0` resolves while the literal `net%d` does not — with the accepted pattern shape matching machinery's own `Validate`. The bare `apiServer:` nit was declined with a rationale and a pinning test, which is a reasonable answer for a pre-existing cosmetic artifact.

The hardening added along the way — mtu digits plus the kernel 68-65535 bounds, the `timeServers` shape guard, and folding six duplicated `extraArgs` loops into `talm.render.component_args` with map/slice rejection — all behaves as documented in local checks.

myasnikovdaniil (approved):

PR #235, opened 2026-09-06
Documentation site at talm.cozystack.io with a generated CLI reference

lexfrei:
DNS for `talm.cozystack.io` resolves, so here is a site to put behind it.

The README was the whole manual: 459 lines, one page, no navigation, no search, no command reference. Sections had also grown into each other, with a 1000-word block on apply-time gates sitting between "gather node information" and "apply config". This splits it into pages under `docs/`, built with MkDocs Material. The prose moves as-is. What changes: heading levels, one cross-reference that pointed at a section now living on another page, and warning blockquotes, which become admonitions. The README keeps the pitch, the features and the install instructions, and links out for the rest.

The CLI reference is generated from the cobra command tree. `TestReferenceDocs` compares the committed pages against that tree, so `go test ./...` fails on drift and `make docs` regenerates them. talm re-exports talosctl's commands, so a Talos bump moves the command surface without anyone touching talm code.

It also drops a false claim that would otherwise have been carried straight into the new docs. talm warned that `--set endpoint=10.0.0.1` reads the dots as key nesting and rendered `{endpoint: {10: {0: {0: 1}}}}`, and told operators to use `--set-string` for IP, CIDR and version literals. strvals does not do that on helm v3 or v4: dots nest only on the left of the `=`, and the value comes through whole. So the warning fired on values that were never at risk and told operators their config was corrupt when it was not. The screen is gone, the flag help now names the real difference (integers and booleans get converted), and two tests pin the parser behaviour so the claim cannot come back.

Before merging, someone has to set the custom domain to `talm.cozystack.io` in Settings > Pages. There is no `CNAME` file, because GitHub ignores one in the artifact when publishing from a workflow, and the domain is not expressible in the workflow either. Until it is set, the 13 absolute links now in the README 404. Provisioning the Pages site itself needs no manual step, the workflow does that through `configure-pages`.

Worth knowing while reviewing:

- `mkdocs build --strict` runs on every PR with anchor, link and nav validation raised to warnings. That workflow has no paths filter on purpose: one skipped by a filter stays Pending and can never be a required check.
- `go-md2man` arrives as an indirect dependency of `cobra/doc`. It stays test-only, so the shipped binary is unchanged.
- The reference test skips on Windows. Flag defaults go through `filepath.Join`, so the pages render with backslashes there.

`cozystack/website` also documents talm under `content/en/docs/*/install/kubernetes/talm.md`, so the two overlap now. Nothing there links to a README anchor, so nothing breaks, but the duplication is worth a follow-up in that repo.


myasnikovdaniil (approved):
APPROVE. I ran `go test ./...` and `mkdocs build --strict`, both pass. I also checked navigation, search and mobile layout in browser. Pages needs to be enabled manually with GitHub Actions as source when setting the custom domain.

PR #237, opened 2026-09-07
Unbreak the lint job under Go 1.27

lexfrei:
The `lint` job fails on every run:

```
panic: file requires newer Go version go1.27 (application built with go1.26)
```

`go-version: stable` now resolves to Go 1.27.1, and the pinned golangci-lint v2.12.2 is built with go1.26, so it dies while typechecking the standard library, before it reaches any code in this repository. The last green `pr.yml` run was 2026-08-04; nothing triggered the workflow between then and Go 1.27 becoming stable, so this went unnoticed.

Bumping the pin to v2.13.2 clears the panic and turns up three things:

- `exhaustruct` is disabled in `.golangci.yml`, but v2.13 ships the v5 rewrite under `exhaustruct_v5` and matches only the new name, so the entry silently stopped applying and 50 findings appeared. Both names are listed now, the way `gomodguard` already handles the same situation.
- The `goconst` half of the nolint directive on `replaceSemanticPaths` no longer suppresses anything, so it is gone.
- `pkg/engine/helm/doc.go` carries upstream's canonical-import-path comment, which Go ignores in module mode. The package is excluded from `modernize` rather than edited, since staying close to upstream is the point of that copy.

The second commit moves the `go` directive to 1.27.1, which is what CI has been building and testing with anyway. Building from source now needs a 1.27 toolchain.

Worth merging before #235, which is blocked by the same red job.

## Summary by CodeRabbit

* **Chores**
  * Updated the Go toolchain language version.
  * Updated the linting workflow to use a newer patch release.
  * Refined linting rules and exclusions to improve compatibility with maintained upstream code.
* **Refactor**
  * Simplified an internal lint directive without changing runtime behavior.


myasnikovdaniil (approved):
APPROVE. I checked the diff and CI results for this commit. Lint and tests pass on Linux and Windows, and the Go 1.27.1 build requirement is documented.

